PR #11, opened 2024-07-25
feat(web): export puppeteer integration

yuyutaotao:


packages/web-integration/src/common/tasks.ts
```diff
@@ -158,7 +158,7 @@ export class PlayWrightActionAgent {
             param: plan.param,
             executor: async (taskParam) => {
               const scrollToEventName = taskParam.scrollType;
-              const innerHeight = await this.page.evaluate(() => window.innerHeight);
+              const innerHeight = await (this.page as PuppeteerPage).evaluate(() => window.innerHeight);
```
zhoushaw, line 161: Consider switching to the executeScript method in cdp

packages/web-integration/src/common/utils.ts
```diff
@@ -30,7 +30,7 @@ export async function parseContextFromPlaywrightPage(
   };
 }
 
-export async function getElementInfosFromPage(page: Page | PlaywrightPage) {
+export async function getElementInfosFromPage(page: WebPage) {
```
zhoushaw, line 33: Since it is referenced by both playwright and puppeteer, the executeScript method is required

PR #62, opened 2024-08-21
fix(html-extract): optimize the logic for extracting HTML content to improve extraction accuracy

zhoushaw:
- Optimize extract logic
- Optimize plan prompt


packages/web-integration/src/extractor/extractor.ts
```diff
@@ -42,7 +50,9 @@ function generateId(numberId: number) {
 
 export function extractTextWithPosition(
   initNode: Node = container,
+  debugModel = false,
 ): ElementInfo[] {
+  setDebugModel(debugModel);
```
yuyutaotao: Debug model -> Debug mode ?

yuyutaotao (approved):

PR #85, opened 2024-09-04
feat(cache): The cache is generalized to support puppeteers and mobile terminals

zhoushaw:
> Summary

* Make caching universal to support mobile and puppeteer

image


packages/web-integration/src/common/agent.ts
```diff
@@ -55,7 +56,7 @@ export class PageAgent {
       executions: [],
     };
     this.taskExecutor = new PageTaskExecutor(this.page, {
-      cache: opts?.cache || { aiTasks: [] },
+      testFilePath: opts?.testFilePath,
```
yuyutaotao: Should we name it as `cacheContext`  or `cacheKey` ? `PageTaskExecutor` should not be coupled with the test file path

packages/web-integration/src/common/utils.ts
```diff
@@ -111,3 +90,68 @@ export function reportFileName(tag = 'web') {
...
+ * @returns The name of the current execution file
+ */
+export function getCurrentExecutionFile(): string {
+  const error = new Error();
+  const stackTrace = error.stack;
+  const pkgDir = findNearestPackageJson(__dirname) || '';
+  if (stackTrace) {
+    const stackLines = stackTrace.split('\n');
+    for (const line of stackLines) {
+      if (line.includes('.spec.') || line.includes('.test.')) {
```
yuyutaotao: What if it is a normal `.ts` file in puppeteer ?

packages/web-integration/src/common/utils.ts
```diff
@@ -111,3 +90,68 @@ export function reportFileName(tag = 'web') {
...
 }
+
+/**
+ * Get the current execution file name
+ * @returns The name of the current execution file
+ */
+export function getCurrentExecutionFile(): string {
+  const error = new Error();
+  const stackTrace = error.stack;
+  const pkgDir = findNearestPackageJson(__dirname) || '';
```
yuyutaotao: Should be something like this
```typescript
const pkgDir = process.cwd();
path.relative(pkgDir,  '/path/to/file/name')
```

yuyutaotao (approved):

PR #136, opened 2024-10-24
feat(ai-model): add claude computer ablity

zhoushaw:


packages/midscene/src/ai-model/prompt/element_inspector.ts
```diff
@@ -1,6 +1,10 @@
 import type { ResponseFormatJSONSchema } from 'openai/resources';
+import { MATCH_BY_POSITION } from '../openai';
 
 export function systemPromptToFindElement() {
+  if (MATCH_BY_POSITION) {
```
yuyutaotao: use `getAIConfig` from 'packages/midscene/src/ai-model/openai/index.ts', so that we can override this config in browser env

packages/midscene/src/ai-model/prompt/element_inspector.ts
```diff
@@ -139,6 +143,36 @@ Output Example:
...
+
+    Please return the result in JSON format as follows:
+
+    \`\`\`json
+    {
+      "elements": [
+        {
+          // Describe the reason for finding this element, replace with actual value in practice
+          "reason": "Reason for finding element 4: It is located in the upper right corner, is an image type, and according to the screenshot, it is a shopping cart icon button",
+          "text": "",
```
yuyutaotao, line 166: should we still keep the 'text' field ? 

If we decide to keep it, could you explain how the AI would use this ‘text’ field?

packages/midscene/src/insight/utils.ts
```diff
@@ -42,6 +43,7 @@ export function writeInsightDump(
   const baseData: DumpMeta = {
     sdkVersion: getVersion(),
     logTime: Date.now(),
+    model_name: process.env[MIDSCENE_MODEL_NAME] || '',
```
yuyutaotao: also use the `getAIConfig` method to get the model name

yuyutaotao (changes requested):

yuyutaotao (approved):

PR #169, opened 2024-11-27
fix: tooltip of env config

ProtosL:
When I tried to use it, I discovered an experience issue right at the beginning, so I fixed it by the way.

QQ_1732694528284


yuyutaotao:
Thanks @ProtosL  ! 
We will ship this fix in the next release of the Chrome extension.

ProtosL:
> Thanks @ProtosL ! We will ship this fix in the next release of the Chrome extension.

Can you merge it? I can't see the merge button. It seems that I don't have permission to merge. Perhaps the repository's permission settings need to be adjusted to allow others to contribute to the Midscene.😃

PR #228, opened 2024-12-31
feat: add bridge mode for extension

yuyutaotao:
Add a bridge mode to the Chrome extension. Developers can use scripts to control the desktop Chrome which they are currently using.

Sample code: https://github.com/web-infra-dev/midscene-example/pull/27


packages/web-integration/tests/unit-test/bridge/agent.test.ts
```diff
@@ -0,0 +1,69 @@
+import {
+  AgentOverChromeBridge,
+  getBridgePageInCliSide,
+} from '@/bridge-mode/agent-cli-side';
+import { describe, expect, it } from 'vitest';
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+describe.skipIf(process.env.CI)(
```
zhoushaw: I believe we ought to relocate this test to the tests/ai/bridge directory.

packages/web-integration/src/bridge-mode/common.ts
```diff
@@ -0,0 +1,32 @@
+export const DefaultBridgeServerPort = 3766;
+export const DefaultLocalEndpoint = `http://127.0.0.1:${DefaultBridgeServerPort}`;
+export const BridgeCallTimeout = 30000;
+export const BridgeCallEvent = 'bridge-call';
```
zhoushaw: Perhaps we could utilize an enum to substitute for listing all of the events.

packages/visualizer/src/init.ts
```diff
@@ -0,0 +1,12 @@
+// biome-ignore lint/style/useNodejsImportProtocol: <explanation>
+import { Buffer } from 'buffer';
+
+window.global ||= window;
```
zhoushaw, line 6: When does this code work?

packages/visualizer/src/extension/bridge.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+const connectTimeout = 30 * 1000;
+const connectRetryInterval = 300;
+export default function Bridge() {
+  const activeBridgePageRef = useRef<ChromeExtensionPageBrowserSide | null>(
+    null,
+  );
+
+  const [bridgeStatus, setBridgeStatus] = useState<
+    'closed' | 'open-for-connection' | 'connected'
```
zhoushaw, line 22: The statuses here can be maintained within an enum in midscene/web/bridge-mode, enabling us to perceive the impact on both sides in the future.

zhoushaw: This status can also be maintained as an enum.

packages/web-integration/src/bridge-mode/agent-cli-side.ts
```diff
@@ -0,0 +1,116 @@
...
+          click: bridgeCaller('mouse.click'),
+          wheel: bridgeCaller('mouse.wheel'),
+          move: bridgeCaller('mouse.move'),
+        };
+        return mouse;
+      }
+
+      if (prop === 'keyboard') {
+        const keyboard: KeyboardAction = {
+          type: bridgeCaller('keyboard.type'),
```
zhoushaw: Could these events be unified into an enum? Otherwise, it's easy to overlook new events in the future.

packages/visualizer/src/init.ts
yuyutaotao: For some strange error like '"global is not defined" in randomBytes, which is introduced by socket.io.

I find this the easiest (maybe not the best) way to solve. 😭

zhoushaw (approved):

PR #295, opened 2025-01-18
docs: add Japanese README

eltociear:
I created Japanese translated README.


yuyutaotao:
Hi @eltociear , 

We don't have a Japanese version of the documents. Therefore, I think the README in Japanese may not be necessary.

PR #296, opened 2025-01-18
Update README.md

krenax:


krenax:
The video included in the README.md is not functional. An updated version would be greatly appreciated.

PR #311, opened 2025-01-22
docs: update wording in index.mdx

reeveng:
Previously, the wording felt slightly awkward, so in this PR, I’ve made some adjustments to improve clarity and flow while retaining the original meaning.


yuyutaotao:
Thanks @reeveng  !

PR #320, opened 2025-01-24
fix: key string capitalization issue

Brass-neck:
image
image


yuyutaotao:
HI @Brass-neck ,
modify this file : packages/web-integration/src/common/tasks.ts , about line 297. 

it should looks like this 

```
function mapToStandardKeyCode(input:string) {
  // ...
}

// ...
            executor: async (taskParam) => {
              assert(taskParam?.value, 'No key to press');
              const  key = mapToStandardKeyCode(taskParam.value);
              assert(key, 'cannot map ....');
              await this.page.keyboard.press(key);
            },
```

PR #518, opened 2025-04-01
fix: force screenshot when the app doesn't allow it

quanru:


packages/android/src/page/index.ts
```diff
@@ -210,6 +216,17 @@ export class Page implements AbstractPage {
     );
   }
 
+  private async forceScreenshot(path: string): Promise<void> {
+    // screenshot which is forbidden by app
+    await this.pushYadb();
```
yuyutaotao: could rename to `this.ensureYadb()`

yuyutaotao (changes requested):

yuyutaotao (approved):

PR #544, opened 2025-04-08
feat(android): implement custom method to retrieve screen size and simplify size handlinga

quanru:


packages/android/src/page/index.ts
```diff
@@ -126,40 +126,45 @@ export class AndroidDevice implements AndroidDevicePage {
...
   }
 
+  private async getScreenSize(): Promise<string> {
+    const adb = await this.getAdb();
+    const stdout = await adb.shell(['wm', 'size']);
+
+    // First try to get Override size
+    let size = new RegExp(/Override size: ([^\r?\n]+)*/g).exec(stdout);
+    if (size && size.length >= 2 && size[1]) {
+      console.log(`Using Override size: ${size[1].trim()}`);
```
yuyutaotao: use getDebug to print debugLogs

yuyutaotao (changes requested):

packages/android/src/page/index.ts
quanru: Is this to prompt the user about which size is currently being used, and do we also need to use debug?

PR #551, opened 2025-04-10
feat: android yaml support

quanru:


packages/core/src/yaml.ts
```diff
@@ -26,7 +28,7 @@ export interface MidsceneYamlTask {
   continueOnError?: boolean;
 }
 
-export interface MidsceneYamlScriptEnv {
+export interface MidsceneYamlScriptWebEnv {
```
yuyutaotao: how about use a common type like 
```
export interface MidsceneYamlScriptEnvBase {
  output:...
}

export interface MidsceneYamlScriptWebEnv extends MidsceneYamlScriptEnvBase {
...
}

File: packages/core/src/yaml.ts [hunk 2]
```diff
@@ -51,6 +53,21 @@ export interface MidsceneYamlScriptEnv {
...
 }
 
+export interface MidsceneYamlScriptAndroidEnv {
+  output?: string;
+  aiActionContext?: string;
+
+  // The Android device ID to connect to, optional, will use the first device if not specified
+  deviceId?: string;
+
+  // The URL or app package to launch, required
```
yuyutaotao: How about making it optional (just use what is currently displayed on the screen)?

yuyutaotao: i mean the `launch` param

yuyutaotao (approved):

PR #631, opened 2025-04-24
feat(core): support custom midscene_run dir

quanru:


packages/core/src/env.ts
```diff
@@ -1,208 +1,81 @@
...
+  OPENAI_USE_AZURE,
+  allConfigFromEnv,
+  getAIConfig,
+  getAIConfigInBoolean,
+  getAIConfigInJson,
+  overrideAIConfig,
+  vlLocateMode,
+} from '@midscene/shared/env';
+
+export {
```
zhoushaw: We can directly re-export

zhoushaw: Perhaps it's also possible to search for exactly how many places have used core/env, as there is also a re-export for this. Could it be considered to directly modify it

packages/shared/src/common.ts
```diff
@@ -1,28 +1,46 @@
...
   if (!isNodeEnv) {
     return '';
   }
 
-  let basePath = path.join(process.cwd(), runDirName);
+  let basePath;
+
+  if (isAbsolute(getMidsceneRunDir())) {
+    basePath = getMidsceneRunDir();
+  } else if (getAIConfig(MIDSCENE_USE_REPORT_TMP_DIR)) {
```
zhoushaw: We can add this variable to the document

zhoushaw: should use getAIConfigInBoolean repalce getAIConfig

yuyutaotao (approved):

PR #651, opened 2025-04-27
WIP: playwright use buffer

quanru:


packages/core/src/utils.ts
```diff
@@ -111,10 +111,28 @@ export function reportHTMLContent(
         '\n</script>'
       );
     });
+
+    // use Buffer to handle large data, avoid string length limit
+    // add newline between each element
+    const newLine = Buffer.from('\n');
```
yuyutaotao: we need a test case for this method

yuyutaotao (approved):

PR #681, opened 2025-04-30
feat(shared): add custom adb path

HBLADEH:
feat(android): add custom adb path

### DIFF
--- a/packages/android/src/page/index.ts
+++ b/packages/android/src/page/index.ts
@@ -1,7 +1,7 @@
 import assert from 'node:assert';
 import fs from 'node:fs';
 import path from 'node:path';
-import type { Point, Size } from '@midscene/core';
+import { getAIConfig, type Point, type Size } from '@midscene/core';
 import type { PageType } from '@midscene/core';
 import { getTmpFile } from '@midscene/core/utils';
 import type { ElementInfo } from '@midscene/shared/extractor';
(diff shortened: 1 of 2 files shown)


yuyutaotao:
Hi @HBLADEH , how about just use the ANDROID_HOME path ?

HBLADEH:
> 你好@HBLADEH，只使用 ANDROID_HOME 路径怎么样？

Sounds good. The original matching method can automatically detect adb.exe under the \platform-tools\ directory of the Android SDK. I think if some users want to use an adb.exe from a specific directory (not within the Android SDK), they can use this method to configure it. (The above reply used translation.)🥺

yuyutaotao:
So does it work if you override ANDROID_HOME before running Midscene SDK?

HBLADEH:
> So does it work if you override ANDROID_HOME before running Midscene SDK?

Sorry, this doesn't seem to have any effect: I set ANDROID_HOME to D:\new_adbpath, but after running it still shows the directory D:\kit\Android\Sdk.

image

yuyutaotao:
It seems that the ANDROID_HOME actually works after some configuration, right?

If you can find a way to set the environmental variable for each process on Windows, then this task can be done.

yuyutaotao:
https://www.perplexity.ai/search/how-to-set-different-env-varia-woxe3DOsQFuxP5DyVZ6uQg

HBLADEH:
> [https://www.perplexity.ai/search/how-to-set- Different-env-varia-woxe3DOsQFuxP5DyVZ6uQg](https://www.perplexity.ai/search/how-to-set-different-env-varia-woxe3DOsQFuxP5DyVZ6uQg)

ok.... thanks for help🥺

yuyutaotao:
This PR has been modified and merged here:  https://github.com/web-infra-dev/midscene/pull/684

PR #688, opened 2025-05-06
feat(android): fix custom adb path no effect

HBLADEH:
Fix the issue where the custom ADB path does not take effect


quanru:
Fix the lints, run npm run lint.

quanru (approved):

PR #699, opened 2025-05-08
fix(core): bundle template content into core package

yuyutaotao:


raiderss:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

zhoushaw (approved):

PR #700, opened 2025-05-09
feat(web-integration): update text element handling to use parent node attributes and content

quanru:


packages/web-integration/tests/unit-test/__snapshots__/web-extractor.test.ts.snap
```diff
@@ -2130,7 +2195,48 @@ exports[`extractor > basic 2`] = `
               "htmlTagName": "<body>",
               "nodeType": "TEXT Node",
             },
-            "content": "phrase "The quick brown fox jumps over the lazy dog" is a well-known English-language pangram, meaning it contains every letter of the alphabet at least once. This sentence has become a standard tool for various applications, particularly     in typing practice, font display, and keyboard testing. The phrase has permeated popular culture and is referenced in various media, including literature and film. Its simplicity and utility have made it a staple in educational contexts and beyond.     For instance, it was famously used as the first message sent over the Moscow–Washington hotline in 19634.",
+            "content": "Data Record
```
zhoushaw, line 2198: It seems abnormal. We should only collect child nodes

PR #714, opened 2025-05-14
fix(android): chinese input compatible with vivo android 14 system

quanru:


yuyutaotao (approved):

quanru:
Fixed by https://github.com/ysbing/YADB/pull/39

PR #793, opened 2025-05-30
fix(android): modify automatically closing the keyboard after typed text

welkang:
主要解决安卓输入文字后会自动关闭键盘，有些页面逻辑识别到键盘关闭，会将页面的下一步操作按钮隐藏，导致流程无法继续下去。

该 PR 将关闭键盘键盘行为移除，后续可以考虑是否改成可配置化。


packages/web-integration/src/chrome-extension/cdpInput.ts
```diff
@@ -35,7 +35,16 @@ export interface KeyDownOptions {
...
+  /**
+   * Time to wait between key presses in milliseconds
+   * @default undefined
+   */
   delay?: number;
+  /**
+   * Whether to hide the keyboard after typing is completed
+   * @default true
+   */
+  hide?: boolean;
```
yuyutaotao: How about rename as `autoDismissKeyboard` ?

welkang: This naming is indeed more semantically clear. I'll make the change.

quanru:
I will continue to support more customization in this MR: #808

PR #823, opened 2025-06-10
feat(mcp): add aiquery support for MCP

RishiKakade:


zhoushaw:
@RishiKakade At present, we have provided the midscene_screenshot method because in our practice, it would be better for the client to read the image directly

PR #853, opened 2025-06-20
Added aiImgTap command (only tested on Android)

tyzamTsing:
Added aiImgTap command (only tested on Android), which can perform template matching operations on icon buttons on the current screen. It is a bit strange that threshold is only about 0.5x, but it can easily reach 0.8~0.9 on Python.
example yaml:
`android:
  deviceId: xxx

tasks:
  - name: to cs page
    flow:
      - aiImgTap: '/private/tmp/cs.png'
        threshold: 0.5
      - sleep: 6000
      - aiAssert: 'AI'`

### DIFF
--- a/packages/core/src/ai-model/common.ts
+++ b/packages/core/src/ai-model/common.ts
@@ -212,6 +212,20 @@ export function adaptBbox(
     return adaptGeminiBbox(bbox, width, height);
   }
 
+  if (vlLocateMode() === 'ecvlmcp') {
+    // ECVLMCP returns pixel coordinates directly, no transformation needed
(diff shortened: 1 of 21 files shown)


yuyutaotao:
Hi @tyzamTsing , is there an introduction to the term 'ECVL'? I only saw some sample MCP server code, but I don’t understand what it is at all.

tyzamTsing:
> Hi @tyzamTsing , is there an introduction to the term 'ECVL'? I only saw some sample MCP server code, but I don’t understand what it is at all.

ECVL is just a generic name.  intend to make a small adjustment on chat for support mcp (android)

PR #854, opened 2025-06-20
feat(report): each time, append the report individually instead of appending in bulk to avoid excessive memory usage

quanru:


quanru:
close #855

PR #950, opened 2025-07-28
fix(shared): Fix hardcode inject-script path

SudoUserReal:
Fix hardcode inject-script path cause error in bundled environment or in non-package.json environment.


EAGzzyCSL:
closed by https://github.com/web-infra-dev/midscene/pull/953

PR #992, opened 2025-08-05
refactor(core): aiAssert use the same implements as aiBoolean

EAGzzyCSL:


packages/core/src/ai-model/prompt/extraction.ts
```diff
@@ -11,10 +11,13 @@ If a key specifies a JSON data type (such as Number, String, Boolean, Object, Ar
 
 If the user provides multiple reference images, please carefully review the reference images with the screenshot and provide the correct answer for <DATA_DEMAND>.
 
+If the user requests reasons to be provided, please provide the thought field in response, less then 100 words.
+
 Return in the following JSON format:
 {
   data: any, // the extracted data. Make sure both the value and scheme meet the DATA_DEMAND. If you want to write some description in this field, use the same language as the DATA_DEMAND.
   errors: [], // string[], error message if any
+  thought: string // the thought process of the extraction, less then 100 words, not required by default.
```
yuyutaotao: order matters for response, should be thought - data - error

PR #1036, opened 2025-08-12
feat(workflow,llm,playwright,puppeteer): and longPress and swipe for playwright and puppeteer

poppyLuo:
feat(workflow,llm,playwright,puppeteer): and longPress and swipe for playwright and puppeteer


yuyutaotao:
merged in #1170

Issue #1116, opened 2025-08-23
[Feature]: Use debugPage instead of console.log for screen size output in AndroidDevice


BeanHsiang:
### What problem does this feature solve?

What problem does this feature solve?
The screen size output inside AndroidDevice currently uses a raw console.log statement instead of the standardized debugPage logger.

Source: web-infra-dev/midscene/packages/android/src/page/index.ts (lines 253–263)

```
console.log(`
DeviceId: ${this.deviceId}
ScreenSize:
${Object.keys(size)
  .filter((key) => size[key as keyof typeof size])
  .map(
    (key) =>
      `  ${key} size: ${size[key as keyof typeof size]}${key === 'override' && size[key as keyof typeof size] ? ' ✅' : ''}`,
  )
  .join('\n')}
`);
```

Why this is an issue:
- Inconsistent logging: The rest of the class uses debugPage for structured, opt‑in debugging.
- Harder to filter: console.log outputs even when users only want targeted debug namespaces.
- Library etiquette: Libraries should avoid unsolicited stdout noise in consuming applications or automation pipelines.
- Missing context control: debugPage likely supports environment-based activation (e.g. DEBUG=android:device), whereas console.log always prints.

Proposed solution:
Replace the console.log block with a debugPage call that formats the same multi-line output.

Optional refinements:
- Collapse into a single-line structured entry (e.g. JSON) when needed.
- Add a secondary concise log: debugPage(`ScreenSize ${this.deviceId} override=${size.override} physical=${size.physical}`);
- Provide a verbose flag (e.g. options.verboseScreenLog) to emit the expanded multi-line format.

Acceptance Criteria:
- console.log usage for this screen size block is removed.
- New debugPage output preserves all existing information (deviceId, override size, physical size, indicator ✅).
- No regression in device connection flow.
- Unit or integration test (if logging is observable) or at least code review confirmation.

Non-goals (for this specific request):
- Refactoring other logging statements.
- Introducing a full logging abstraction beyond existing debugPage usage.

Impact:
Cleaner logs for users integrating midscene into CI, automated test harnesses, and multi-device orchestration systems.

### What does the proposed API look like?

Replace the console.log block in getScreenSize() with debugPage for outputting screen size and device information.

Source file (for the snippet): web-infra-dev/midscene/packages/android/src/page/index.ts (lines 253–263)

Current snippet:
```
console.log(`
DeviceId: ${this.deviceId}
ScreenSize:
${Object.keys(size)
[...]

PR #1129, opened 2025-08-26
feat(core): rename page -> interface

yuyutaotao:


yuyutaotao:
#1116 will be fixed in this PR

PR #1283, opened 2025-10-09
feat(android): change default ime strategy

quanru:
## Summary
- hoist the IME strategy string literals into shared constants at the top of the module
- reuse the constants for type declarations, defaults, and comparisons

## Testing
- npm test


yuyutaotao:
The explanation of the imeStrategy parameter in the document needs further optimization and additional information: `yadb-for-non-ascii` means use yadb for non-ascii words input, while `always-yadb` means always use yadb to do input tasks. Try switching this config when it fails to input texts.

yuyutaotao:
@copilot The explanation of the imeStrategy parameter in the document needs further optimization and additional information: yadb-for-non-ascii means use yadb for non-ascii words input, while always-yadb means always use yadb to do input tasks. Try switching this config when it fails to input texts.

Please help to modify the code and commit

PR #1460, opened 2025-11-15
feat(core): add configurable QWEN high resolution images support

AdityaNarayan05:
- Add MIDSCENE_USE_QWEN_HIGH_RES environment variable for controlling vl_high_resolution_images parameter
- Support three states: enabled (true), disabled (false), unspecified (defaults to enabled)
- Only affects QWEN models (qwen-vl, qwen3-vl), ignored for other model types
- Add comprehensive test coverage for all configuration scenarios
- Implement case-insensitive parsing for extended truthy/falsy values

### DIFF
--- a/packages/core/src/ai-model/service-caller/index.ts
+++ b/packages/core/src/ai-model/service-caller/index.ts
@@ -225,7 +225,7 @@ export async function callAI(
         : Number.parseInt(maxTokens || '2048', 10),
     ...(vlMode === 'qwen-vl' || vlMode === 'qwen3-vl' // qwen specific config
       ? {
-          vl_high_resolution_images: true,
+          ...(modelConfig.qwenHighResolution === false ? {} : { vl_high_resolution_images: true }),
         }
       : {}),
   };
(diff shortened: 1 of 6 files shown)


yuyutaotao:
Hi @AdityaNarayan05 , In which cases should you remove the resolution parameter?

Actually in the coming v1.0 version, we will remove this for qwen3-vl. But it's necessary for qwen2.5-vl according to the document of alibaba cloud.

AdityaNarayan05:
Hi @yuyutaotao,

Thank you for the feedback! You raise an excellent point about the different QWEN 
model versions.

Current use cases for removing the resolution parameter:
1. Performance optimization - When high resolution processing isn't needed for the 
specific use case
2. Cost reduction - High resolution processing may incur higher API costs
3. Compatibility issues - Some deployment environments might have issues with the 
parameter
4. User preference - Allow users to opt-out if they prefer faster processing over 
higher accuracy

Regarding v1.0 and model-specific behavior:
Since you mentioned that qwen3-vl will have this removed in v1.0 but qwen2.5-vl still
needs it, should we update the implementation to be more model-specific? 

Proposed approach:
Instead of applying this to all QWEN models, we could:
1. Only apply the parameter to qwen2.5-vl (and other models that need it)
2. Exclude qwen3-vl from this configuration
3. Make it future-proof for v1.0 changes

Would you like me to update the implementation to be more granular based on specific 
model versions? I can modify the code to check for exact model names rather than just
the qwen prefix.

Let me know your preference and I'll make the necessary adjustments!

PR #1529, opened 2025-12-02
feat(core): add AI API timeout configuration support

quanru:
## Summary

- Add `MIDSCENE_MODEL_TIMEOUT` environment variable for configuring AI API call timeout
- Add `AgentOpt.modelTimeout` parameter for JavaScript/TypeScript configuration
- Timeout value is in milliseconds, defaults to OpenAI SDK default (10 minutes)

## Configuration Priority

1. `AgentOpt.modelTimeout` (highest priority)
2. `MIDSCENE_MODEL_TIMEOUT` environment variable
3. OpenAI SDK default (10 minutes / 600000ms)

### Environment Variable

```bash
# Set 60 second timeout
export MIDSCENE_MODEL_TIMEOUT=60000
```

### JavaScript Configuration

```typescript
const agent = new Agent(page, {
  // Option 1: Direct parameter
  modelTimeout: 60000, // 60 seconds

  // Option 2: Via modelConfig
  modelConfig: {
    MIDSCENE_MODEL_TIMEOUT: 60000,
  }
});
```

## Changes

- `packages/shared/src/env/types.ts` - Add `MIDSCENE_MODEL_TIMEOUT` constant and `timeout` field to `IModelConfig`
- `packages/core/src/types.ts` - Add `modelTimeout` property to `AgentOpt` interface
- `packages/shared/src/env/model-config-manager.ts` - Resolve timeout from multiple configuration sources
- `packages/core/src/agent/agent.ts` - Pass `modelTimeout` to `ModelConfigManager`
- `packages/core/src/ai-model/service-caller/index.ts` - Apply timeout to OpenAI client initialization
- `apps/site/docs/en/model-config.mdx` - Update English documentation
- `apps/site/docs/zh/model-config.mdx` - Update Chinese documentation

## Test Plan

- [x] Unit tests pass (`npx nx test @midscene/core`)
- [x] Lint passes (`npm run lint`)
- [x] Verified timeout configuration works with all three methods

### DIFF
--- a/apps/site/docs/en/model-config.mdx
+++ b/apps/site/docs/en/model-config.mdx
@@ -240,9 +240,6 @@ const agent = new Agent(page, {
   modelConfig: {
     MIDSCENE_MODEL_TIMEOUT: '60000', // 60 seconds
     MIDSCENE_MODEL_NAME: 'qwen3-vl-plus',
-    // For per-intent timeout configuration:
-    MIDSCENE_INSIGHT_MODEL_TIMEOUT: '90000',
-    MIDSCENE_PLANNING_MODEL_TIMEOUT: '120000',
     // ... other configurations
   }
 });
--- a/apps/site/docs/zh/model-config.mdx
+++ b/apps/site/docs/zh/model-config.mdx
@@ -244,9 +244,6 @@ const agent = new Agent(page, {
   modelConfig: {
     MIDSCENE_MODEL_TIMEOUT: '60000', // 60 秒
     MIDSCENE_MODEL_NAME: 'qwen3-vl-plus',
-    // 为每个意图单独配置超时时间：
-    MIDSCENE_INSIGHT_MODEL_TIMEOUT: '90000',
-    MIDSCENE_PLANNING_MODEL_TIMEOUT: '120000',
     // ... 其他配置
   }
 });


quanru:
close https://github.com/web-infra-dev/midscene/issues/873

packages/shared/src/env/model-config-manager.ts
```diff
@@ -44,6 +47,14 @@ export class ModelConfigManager {
       configMap = this.globalConfigManager?.getAllEnvConfig() || {};
     }
 
+    // Resolve timeout priority (modelTimeout always takes precedence if provided):
```
yuyutaotao: ⁠We should add two more variables such as `MIDSCENE_PLANNING_MODEL_TIMEOUT` and `MIDSCENE_INSIGHT_MODEL_TIMEOUT`, therefore there is no need for a special "resolvedTimeout".

yuyutaotao (changes requested):

packages/shared/src/env/model-config-manager.ts
```diff
@@ -22,13 +22,16 @@ export class ModelConfigManager {
 
   private modelConfig?: TModelConfig;
   private createOpenAIClientFn?: CreateOpenAIClientFn;
+  private modelTimeout?: number;
 
   constructor(
     modelConfig?: TModelConfig,
     createOpenAIClientFn?: CreateOpenAIClientFn,
+    modelTimeout?: number,
```
yuyutaotao: I think the `modelTImeout` has already been included in the `modelConfig` with the name `timeout` ?

packages/core/src/types.ts
```diff
@@ -637,6 +637,12 @@ export interface AgentOpt {
    */
   replanningCycleLimit?: number;
 
+  /**
+   * Timeout for AI model API calls in milliseconds.
+   * If not set, uses MIDSCENE_MODEL_TIMEOUT env var or OpenAI SDK default (10 minutes).
+   */
+  modelTimeout?: number;
```
yuyutaotao: this is also unnecessary

packages/shared/src/env/model-config-manager.ts
quanru: If using modelConfig, the user has to construct the entire configuration, but they might only need to set the timeout duration.

PR #1550, opened 2025-12-04
fix(core): Fix HTTP/SOCKS proxy configuration with OpenAI SDK v6.3.0 fetch API

quanru:
## Summary

Fixes proxy configuration to work correctly with OpenAI SDK v6.3.0, which uses undici-based fetch API instead of traditional Node.js HTTP modules.

## Problem

The previous implementation used `httpAgent` parameter, which was being logged as "using http proxy" but was actually **ignored** by OpenAI SDK v6.3.0. The SDK uses fetch API (powered by undici) and requires `fetchOptions.dispatcher` instead.

### Core Changes
- **HTTP Proxy**: Replaced `httpAgent` parameter with `fetchOptions.dispatcher` using `undici.ProxyAgent`
- **SOCKS Proxy**: Added SOCKS4/SOCKS5 support using `fetch-socks` package with `socksDispatcher`
- **Authentication**: Both HTTP and SOCKS proxies support credentials in URL format
- **Dynamic Imports**: Used string variable imports to prevent bundlers from including Node.js-only modules in browser builds
- **Browser Compatibility**: Added environment checks with appropriate warnings for browser environments

### Dependencies
- Added `undici ^6.0.0` for HTTP proxy support
- Added `fetch-socks ^1.3.0` for SOCKS proxy support

### Browser Build Fixes
- Added alias configuration in `apps/playground/rsbuild.config.ts` to exclude `undici` and `fetch-socks` from browser builds
- Prevents "Module not found: Can't resolve 'worker_threads'" errors

### Testing
- Converted manual test scripts to comprehensive unit tests
- Tests verify proxy configuration by expecting connection attempts to configured proxy servers
- Coverage: HTTP proxy, SOCKS proxy (v4 & v5), authentication, priority, error handling

## Verification

All unit tests pass (9/9):
- ✅ HTTP proxy configuration using undici ProxyAgent
- ✅ SOCKS5 proxy configuration
- ✅ SOCKS4 proxy URL parsing
- ✅ Proxy authentication (credentials in URL)
- ✅ HTTP proxy priority over SOCKS
- ✅ Invalid SOCKS URL validation
- ✅ Direct connection without proxy

### Before (Broken)
```typescript
const openAIOptions = {
  httpAgent: proxyAgent,  // ❌ Ignored by SDK
  ...
};
```

### After (Working)
```typescript
const { ProxyAgent } = await import('undici');
const openAIOptions = {
  fetchOptions: {
    dispatcher: proxyAgent  // ✅ Correct for fetch API
  },
  ...
};
```

## Related Issues

Fixes the issue where proxy configuration was being logged but not actually applied to API requests.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/core/src/ai-model/service-caller/index.ts
+++ b/packages/core/src/ai-model/service-caller/index.ts
@@ -50,10 +50,18 @@ async function createChatClient({
   const debugProxy = getDebug('ai:call:proxy');
 
   // Helper function to sanitize proxy URL for logging (remove credentials)
+  // Uses URL API instead of regex to avoid ReDoS vulnerabilities
   const sanitizeProxyUrl = (url: string): string => {
     try {
-      return url.replace(/\/\/([^:]+):([^@]+)@/, '//$1:****@');
+      const parsed = new URL(url);
+      if (parsed.username) {
+        // Keep username for debugging, hide password for security
+        parsed.password = '****';
+        return parsed.href;
+      }
+      return url;
     } catch {
+      // If URL parsing fails, return original URL (will be caught later)
       return url;
     }
   };


packages/core/package.json
```diff
@@ -76,8 +76,10 @@
     "@ui-tars/action-parser": "1.2.3",
     "dayjs": "^1.11.11",
     "dotenv": "^16.4.5",
+    "fetch-socks": "^1.3.0",
```
yuyutaotao, line 79: Can we remove the `https-proxy-agent` from the deps now ?

yuyutaotao (changes requested):

packages/core/package.json
quanru: done

PR #1607, opened 2025-12-17
runYaml do not replace variables

dorahappy:


yuyutaotao:
Let's use this one instead: https://github.com/web-infra-dev/midscene/pull/1609/changes

PR #1647, opened 2025-12-23
Add Playwright based MCP server

lilac:
In v1 the MCP server for web automation only supports bridge mode. This change bring the deprecated MCP server back.

### DIFF
--- a/packages/web-playwright-mcp/src/playwright-tools.ts
+++ b/packages/web-playwright-mcp/src/playwright-tools.ts
@@ -51,6 +51,10 @@ export class PlaywrightMidsceneTools extends BaseMidsceneTools<PlaywrightAgent> 
     return this.agent;
   }
 
+  // Maximum viewport dimensions
+  private static readonly MAX_VIEWPORT_WIDTH = 1280;
+  private static readonly MAX_VIEWPORT_HEIGHT = 720;
+
   /**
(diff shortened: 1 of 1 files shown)


quanru:
Thanks for the contribution! However, our current codebase already supports non-bridge mode for the web MCP server.

In `packages/web-integration/src/cli.ts`, the default mode is **Puppeteer-based** (non-bridge). The `--bridge` flag is required to use bridge mode:

```typescript
const isBridge = process.argv.includes('--bridge');
const tools = isBridge
  ? new WebMidsceneTools()           // Bridge mode
  : new WebPuppeteerMidsceneTools(); // Default: Puppeteer non-bridge mode
```

The `WebPuppeteerMidsceneTools` class (in `packages/web-integration/src/mcp-tools-puppeteer.ts`) provides direct browser automation via Puppeteer without requiring the Chrome extension bridge. It launches a persistent Chrome instance and supports `web_connect`, `web_disconnect`, and `web_close` operations - similar to what this PR proposes with Playwright.

So we don't think adding a separate Playwright-based MCP package is necessary at this point, as the existing Puppeteer mode already covers this use case. If you have specific requirements that Puppeteer mode doesn't satisfy, feel free to share them and we can discuss further.

quanru:
If there is still a need, please reopen the current pr

PR #1709, opened 2025-12-31
feat(web-integration): implement file upload functionality with tests and HTML fixture

quanru:


packages/core/src/agent/agent.ts
```diff
@@ -588,14 +587,30 @@ export class Agent<
...
 
     const detailedLocateParam = buildDetailedLocateParam(locatePrompt, opt);
+    const fileChooserInterface = hasFileChooserCapability(this.interface)
+      ? this.interface
+      : null;
 
-    return this.callActionInActionSpace('Tap', {
-      locate: detailedLocateParam,
-    });
+    if (opt?.files && fileChooserInterface) {
```
yuyutaotao: should print a warning if `opt.files` is set but fileChooseInterface is not available

packages/web-integration/src/web-page.ts
```diff
@@ -429,6 +429,8 @@ export const commonWebActionsForWebPage = <T extends AbstractWebPage>(
   defineActionTap(async (param) => {
     const element = param.locate;
     assert(element, 'Element not found, cannot tap');
+
+    // Pure tap action - file handling is done at Page layer via setFileChooserHandler
```
yuyutaotao, line 433: this comment can be removed now

PR #1749, opened 2026-01-09
fix: resolve MIDSCENE_MODEL_MAX_TOKENS environment variable not working

### LINKED ISSUE #1748: [Bug]: Environment variable MIDSCENE_MODEL_MAX_TOKENS not applied due to type check failure
### Version

```sh
System:
  - OS: macOS 15.7 (Darwin 24.6.0)
  - Architecture: arm64

  Browsers:
  - N/A (Backend issue)

  npm packages:
  - @midscene/core: 1.0.4 / 1.1.0
  - Node.js: v20.19.2
  - npm: 10.8.2
```

### Details

 Description

  The environment variable MIDSCENE_MODEL_MAX_TOKENS is not being applied correctly when set in the .env file or system environment. This causes AI model API calls to use the default token limit instead of the configured value, potentially leading to "content empty" errors and response truncation as follow.

> failed to call AI model service (gemini-3-pro): empty content
> Trouble shooting: https://midscenejs.com/model-provider.html
> Error: failed to call AI model service (gemini-3-pro): empty content 

Image

 ## Root Cause

  Environment variables in Node.js are always of type string, but the code performs a type check typeof maxTokens === 'number' which causes the value to be discarded:
```
  File: @midscene/core/dist/es/ai-model/service-caller/index.mjs

  Current code (line ~131):
  const maxTokens = globalConfigManager.getEnvConfigValue(MIDSCENE_MODEL_MAX_TOKENS) ??
                    globalConfigManager.getEnvConfigValue(OPENAI_MAX_TOKENS);

  // Type check fails for string values
  const commonConfig = {
    // ...
    max_tokens: 'number' == typeof maxTokens ? maxTokens : void 0  // Returns undefined for "16000"
  };
```
 ##  Issue:
  - When MIDSCENE_MODEL_MAX_TOKENS=16000 is set, maxTokens equals "16000" (string)
  - Type check 'number' == typeof maxTokens returns false
  - Result: max_tokens: undefined is passed to the API
  - API uses its default token limit (often lower than needed)

### Reproduce link

needless

### Reproduce Steps

 Reproduce Steps

  1. Set environment variable:
   In .env file
  MIDSCENE_MODEL_MAX_TOKENS=16000
  2. Add logging to verify the issue:
  // In @midscene/core/dist/es/ai-model/service-caller/index.mjs
  console.log('maxTokens value:', maxTokens);
  console.log('maxTokens type:', typeof maxTokens);
  console.log('Actual max_tokens:', commonConfig.max_tokens);
  3. Run any Midscene test:
  npm run test
  4. Observe the logs:
  maxTokens value: 16000         // ✅ Value is read
  maxTokens type: string         // ❌ Type is string, not number
  Actual max_tokens: undefined   // ❌ Not applied to API config
  5. Check API response:
    - If response is large, may see finish_reason: "length" (token limit reached)
    - Response may be truncated

  Expected Behavior

  - MIDSCENE_MODEL_MAX_TOKENS=16000 should be applied to API calls as max_tokens: 16000
  - The type check should handle string-to-number conversion for environment variables

  Actual Behavior

  - Environment variable is read but discarded due to type check
  - API calls use max_tokens: undefined (default limit)
  - May cause unexpected token limit errors

wujunwei:
The environment variable MIDSCENE_MODEL_MAX_TOKENS was not being applied because getEnvConfigValue() returns string | undefined, but the code checked `typeof maxTokens === 'number'` which would never be true.

Changed to convert the string value to number using Number().

Fix #1748

### DIFF
--- a/packages/core/src/ai-model/service-caller/index.ts
+++ b/packages/core/src/ai-model/service-caller/index.ts
@@ -220,8 +220,8 @@ export async function callAI(
     });
 
   const maxTokens =
-    globalConfigManager.getEnvConfigValue(MIDSCENE_MODEL_MAX_TOKENS) ??
-    globalConfigManager.getEnvConfigValue(OPENAI_MAX_TOKENS);
+    globalConfigManager.getEnvConfigValueAsNumber(MIDSCENE_MODEL_MAX_TOKENS) ??
+    globalConfigManager.getEnvConfigValueAsNumber(OPENAI_MAX_TOKENS);
   const debugCall = getDebug('ai:call');
   const debugProfileStats = getDebug('ai:profile:stats');
   const debugProfileDetail = getDebug('ai:profile:detail');
@@ -258,7 +258,7 @@ export async function callAI(
   const commonConfig = {
     temperature,
     stream: !!isStreaming,
-    max_tokens: typeof maxTokens === 'number' ? maxTokens : undefined,
+    max_tokens: maxTokens,
     ...(vlMode === 'qwen2.5-vl' // qwen vl v2 specific config
       ? {
           vl_high_resolution_images: true,
--- a/packages/shared/src/env/global-config-manager.ts
+++ b/packages/shared/src/env/global-config-manager.ts
@@ -118,6 +118,20 @@ export class GlobalConfigManager {
     return !!value.trim();
   }
 
+  /**
+   * Read string environment variable value and convert it to number.
+   * Returns undefined if the value is not set or cannot be converted to a valid number.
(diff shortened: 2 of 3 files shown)


yuyutaotao:
@wujunwei  How about this:

* implement a new  `getEnvConfigValueAsNumber` which returns `number | undefined`
* remember to add test cases

wujunwei:
> @wujunwei How about this:
> 
> * implement a new  `getEnvConfigValueAsNumber` which returns `number | undefined`
> * remember to add test cases

Done !

PR #1760, opened 2026-01-12
fix(shared): add YAML indentation rules to prevent LLM generating incorrect format

quanru:
## Summary
- Add explicit YAML indentation rules section to `YAML_EXAMPLE_CODE` constant
- Clarify that parameters (direction, scrollType, locate) must be sibling keys aligned with the action key, not nested children
- Include correct and incorrect examples to guide LLM in generating proper YAML format

## Problem
The recorder was generating YAML with incorrect indentation for actions like `aiScroll`. Parameters were being indented 2 extra spaces as nested children instead of being aligned as sibling keys.

**Expected format:**
```yaml
      - aiScroll:
        direction: down
        scrollType: singleAction
```

**Incorrect format being generated:**
```yaml
      - aiScroll:
          direction: down
          scrollType: singleAction
```

## Test plan
- [ ] Verify lint passes
- [ ] Test recorder YAML generation to confirm correct indentation

### DIFF
--- a/packages/shared/src/constants/example-code.ts
+++ b/packages/shared/src/constants/example-code.ts
@@ -113,6 +113,18 @@ tasks:
 - Add deepThink: true for complex interactions
 - Keep task names concise but descriptive
 
+4. CRITICAL - YAML Indentation Rules:
+- For actions with additional parameters (aiScroll, aiInput, aiKeyboardPress), the parameters must be SIBLING keys, NOT nested children
+- Parameters like direction, scrollType, locate must align with the action key, not indented further
+- CORRECT indentation example:
+      - aiScroll:
+        direction: down
+        scrollType: singleAction
+- WRONG indentation (DO NOT do this):
+      - aiScroll:
+          direction: down
+          scrollType: singleAction
+
 
 
 YAML type


quanru:
Thanks for the review! However, I need to respectfully disagree with Copilot's suggestions here.

**The indentation I proposed is correct.** Let me explain with evidence:

### 1. YAML Parsing Verification

```yaml
# Format A (what I marked as CORRECT):
- aiScroll:
  direction: down
  scrollType: singleAction
```
Parses to: `[{'aiScroll': None, 'direction': 'down', 'scrollType': 'singleAction'}]`

```yaml
# Format B (what I marked as WRONG):
- aiScroll:
    direction: down
    scrollType: singleAction
```
Parses to: `[{'aiScroll': {'direction': 'down', 'scrollType': 'singleAction'}}]`

### 2. Code Evidence from player.ts (lines 435-438)

```typescript
// Compatibility with previous version:
// Old format: { aiScroll: null, locate?: TUserPrompt, direction, scrollType, distance? }
// New format - 1: { aiScroll: TUserPrompt, direction, scrollType, distance? }
// New format - 2: { aiScroll: undefined, locate: TUserPrompt, direction, scrollType, distance? }
```

The code explicitly expects `direction` and `scrollType` to be **sibling properties** of `aiScroll` in the same object, NOT nested inside `aiScroll`'s value.

### 3. The Actual Bug Report

The user reported that the recorder was generating:
```yaml
- aiScroll:
    scrollType: 'singleAction'  # 10 spaces - WRONG
```

When it should generate:
```yaml
- aiScroll:
  scrollType: 'singleAction'    # 8 spaces - CORRECT
```

The extra 2-space indentation causes `scrollType` to become a nested child of `aiScroll` instead of a sibling key, which breaks the YAML player parsing.

---

**Summary:** Copilot's suggestion would actually introduce the exact bug we're trying to fix. The terminology "sibling keys" may not be standard YAML parlance, but the indentation examples are correct for how Midscene's YAML parser expects the data structure.

yuyutaotao (approved):

PR #1799, opened 2026-01-16
chore(docs): doc for auto-glm

EAGzzyCSL:


wujunwei:
Will the next version support this feature?

PR #1911, opened 2026-02-02
fix(web-integration): remove unnecessary 100px height adjustment in headed mode

quanru:
## Summary
- Remove the hardcoded 100px height offset for address bar compensation in headed mode
- This adjustment was unreliable across different browsers, systems, and scaling factors
- Since headed mode already uses `defaultViewport: null` to auto-fit the viewport, this adjustment is unnecessary

## Related
- Related to PR #1908 which copies this logic - if merged, that PR should also remove this adjustment


quanru:
Changes merged into PR #1908

PR #2112, opened 2026-03-10
feat(all): reportMergingTool support single report

frank-mupt:


quanru (commented):
Main concern: the new `reportFilePath?: string` flow is only implemented in `ReportMergingTool`, but it is not wired through the report viewer yet.

1. `mergeReports()` now writes `<script type="midscene_web_dump">` entries with an empty body when `reportFilePath` is missing. However, the viewer still treats every dump script as JSON and eagerly selects the first case on load. If the first case is skipped, `dumps[0].get()` falls back to a non-`GroupedActionDump` object and `setGroupedDump()` still reads `dump.executions.length`, which can crash the page before the new disabled-click guard has any chance to help.

2. Related to that, the `all skipped` case is still broken in the viewer path. The new unit tests only assert that the merged HTML contains the expected attributes, but they do not verify that `apps/report` can actually load and render those merged reports.

3. This also broadens the public type too much. Making `reportFilePath` optional for all `ReportFileWithAttributes` values allows invalid states such as `passed + undefined reportFilePath`. A narrower model would be safer here, for example a discriminated union for skipped/non-viewable cases instead of weakening the whole API.

Minor point: removing the old `<= 1` guard also changes the `0 reports` behavior. `mergeReports()` now creates and returns an empty template-only report when nothing was appended. I think supporting `1 report` makes sense, but `0 report` should probably still return `null` or throw explicitly.

PR #2121, opened 2026-03-11
fix cache flow is empty && runAdbShell not available

e790a8:
启用缓存发现，aiAct只有Plan时，生成的缓存中flow为[]，导致二次运行使用缓存时，这个步骤就会空跑，没有发挥其作用。
修复：aiAct生成缓存保存时 && 读取Plan缓存使用时，判断是否有值
image
image

### DIFF
--- a/packages/core/src/yaml/player.ts
+++ b/packages/core/src/yaml/player.ts
@@ -100,7 +100,7 @@ const isStringParamSchema = (schema?: ZodTypeAny): boolean => {
   }
 };
 
-const buildSpecialStringActionParam = (
+const buildLaunchOrAdbShellParam = (
   actionName: string,
   interfaceAlias: string | undefined,
   value: string,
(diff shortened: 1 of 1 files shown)


e790a8:
补充：YAML 脚本中 runAdbShell 不会执行，按照文档中说明的格式编写的
image

quanru:
Follow-up fix pushed in `b5f926d04`.

This addresses the two issues found during review:

- Plan cache entries with empty/invalid `yamlWorkflow` are now treated as unusable matches instead of plain cache misses, so a re-plan will overwrite the stale record instead of appending duplicates forever.
- YAML string fallback for `Launch` / `RunAdbShell` now wraps to the real action params (`{ uri }` / `{ command }`) when helper methods are unavailable, instead of falling through to the `locate` branch.

Also updated regression coverage in `core`, `android`, and `web-integration` to cover the real object-schema path and the stale-cache overwrite path.

Verified with:
- `packages/core`: `player-action-dispatch`, `task-cache-empty-flow`, `aiaction-cacheable`
- `packages/android`: `yaml-runAdbShell`
- `packages/web-integration`: `task-cache`

PR #2133, opened 2026-03-12
feat(core): add codex app-server provider

ScriptedAlchemy:
## Summary
- add Codex CLI `app-server` as a Midscene model provider so a local Codex OAuth session can be used without a separate API key
- document the `codex://app-server` configuration flow in the English and Chinese model config docs
- add unit coverage for provider detection, reasoning-effort mapping, and Codex turn payload conversion

## Problem
Midscene currently assumes model access comes from an OpenAI-compatible HTTP endpoint plus API key. That creates unnecessary friction for users who already authenticate through Codex CLI and want to reuse that local OAuth session. The Codex adapter path also needs to consume turn notifications and inline server requests correctly; otherwise notifications can be re-queued and a run can stall instead of completing cleanly.

## Impact
- enables Midscene users to set `MIDSCENE_MODEL_BASE_URL="codex://app-server"` and reuse `codex login`
- prevents Codex-backed runs from hanging when the app-server emits notifications or `requestUserInput` events during a turn
- gives docs readers a supported Codex configuration path in both languages

## Validation
- added unit coverage in `packages/core/tests/unit-test/codex-app-server-provider.test.ts`
- not run locally in this session


ScriptedAlchemy:
Superseded by #2134, which uses the same branch name pushed directly to web-infra-dev/midscene instead of the fork-backed head branch.

Issue #2179, opened 2026-03-19
[Bug]: aiNumber extraction succeeds in report but still throws an error


OneSir:
### Version

```sh
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (12) x64 Intel(R) Xeon(R) Gold 6132 CPU @ 2.60GHz
    Memory: 20.93 GB / 28.00 GB
    Container: Yes
    Shell: 4.3.48 - /bin/bash
  npmPackages:
    @midscene/web: ^1.5.6 => 1.5.6
```

### Details
The report shows the value was successfully extracted, but the final result still throws an error/failure.

### Reproduce Steps
```js
await aiNumber(`Extract the numeric value of '访客数' (visitor count)`);
```
According to the report, the extraction was successful, but it still ended up failing with an error.

---

### Details

![Image](https://github.com/user-attachments/assets/c1358d01-1617-45cd-ab83-faf4e39d0b30)

### Reproduce link

暂无

### Reproduce Steps

await aiNumber(`提取'访客数'的数值`);
看报告已经提取成功了，但最后还是异常失败了

PR #2185, opened 2026-03-19
fix(aiNumber):To ensure proper matching with the structure returned by the AI

xianglongfei-8888:
issues: https://github.com/web-infra-dev/midscene/issues/2179

Root cause: The code uses a fixed keyOfResult='result 'to read the return result, but the key name in the actual result returned by AI is Number/String/Boolean (consistent with the type parameter), resulting in the read result being undefined and throwing a "No result in query data" error, but in reality, AI has correctly extracted the value.

Fix: When setting up a query request for Number/String/Boolean types, also set the keyOfResult to the same value as the type to ensure proper matching with the structure returned by the AI.

### DIFF
--- a/packages/core/src/agent/tasks.ts
+++ b/packages/core/src/agent/tasks.ts
@@ -594,6 +594,7 @@ export class TaskExecutor {
             [keyOfResult]: booleanPrompt,
           };
         } else if (ifTypeRestricted) {
+          keyOfResult = type;
           demandInput = {
             [keyOfResult]: `${type}, ${demand}`,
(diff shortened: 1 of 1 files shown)


OneSir:
试了一下，确实解决了👍🏻

xianglongfei-8888:
> 试了一下，确实解决了👍🏻

@OneSir  需要管理员审核，才能合入哈

xianglongfei-8888:
@quanru 帮忙审核下代码，谢谢！

PR #2191, opened 2026-03-20
feat(report): add advanced e2e tests with merged report and error cases

quanru:
## Summary
- Extend `generate-demo-report.mjs` to run both a passing and a failing Midscene test, then merge them via `ReportMergingTool` into `demo-merged.html`, enabling multi-case e2e testing.
- Add 7 new e2e test suites:
  - **multi-case-selector**: case selector dropdown, switching between cases, filtering by status
  - **test-status-display**: verify Total/Passed/Failed summary cards show correct counts
  - **error-detail**: switch to failed case, click failed task, verify error info in detail panel
  - **dark-mode**: toggle dark/light theme
  - **timeline-interaction**: click timeline screenshot to navigate, collapse/expand timeline
  - **detail-panel-views**: screenshot + info panel display, Tab key to switch to JSON view
  - **model-call-details**: toggle Model Call Details checkbox, verify token columns appear/disappear

## Validation
- `pnpm run lint` from repo root
- `pnpm run e2e` in `apps/report` (pending CI)


quanru:
**Re: P1 — Decouple report e2e from live SauceDemo generation**

This is by design — the user explicitly requested generating reports from real test runs rather than using static test data. SauceDemo is a stable public test site maintained by Sauce Labs, and other e2e tests in this repo (`packages/cli/tests/midscene_scripts/online/online.yaml`) already depend on it. If offline support becomes necessary, we can add a fallback to pre-generated reports later.

**Re: P2 — Don't swallow failures from the supposed passing report**

Good catch — fixed. `runYamlAndFindReport` now accepts a `tolerateFailure` parameter. Only the intentionally-failing report (`generate-report-failed.yaml`) uses `tolerateFailure=true`. The passing report will throw on errors as expected.

PR #2214, opened 2026-03-24
feat(android, harmony): add terminate action to support force-stopping apps and update documentation

zifengjiang:
Add a terminate API for both Android and Harmony Agent.

Register a Terminate action in the device action space, enabling force-stopping applications by package name or bundle name.

Add unit tests for agent/device, covering success cases, URI interception, and failure branches.


zifengjiang:
@quanru 幸苦处理下

quanru (approved):

PR #2279, opened 2026-04-02
fix(web-integration): align yaml waitForNetworkIdle behavior and docs

EAGzzyCSL:


EAGzzyCSL:
beta：https://github.com/web-infra-dev/midscene/actions/runs/23937425983

JiwaniZakir (commented):
The new test in `agent-launcher.test.ts` verifies the plumbing by inspecting the internal property `(agent.page as any).waitForNetworkIdleTimeout`, which is testing implementation state rather than observable behavior — the `base-page-invoke-action.test.ts` test is the more meaningful one here since it actually asserts `waitForNetworkIdle` is called with the right `timeout` argument. More importantly, neither the new agent-launcher test nor any other test in this diff covers the `continueOnNetworkIdleError: false` code path during the initial `goto` — the docs now explicitly call out that this setting causes a hard failure on timeout during URL load, but that behavior appears entirely untested. In `agent-launcher.ts`, the guard `typeof target.waitForNetworkIdle?.timeout === 'number'` correctly avoids passing stringified values, but a `timeout: 0` would silently be forwarded and could cause every action-time idle wait to immediately error out, which may be worth clamping or documenting.

quanru (approved):

PR #2337, opened 2026-04-10
perf: cache resolved system Chrome path

vincerevu:
What:
Introduced a cachedSystemChromePath variable to cache the resolved path returned by getSystemChromePath. Added a clearSystemChromePathCache function to reset the cache between test cases.

 Why:
getSystemChromePath uses node:fs existsSync checks synchronously across an array of possible paths. Repeated calls hit the filesystem unnecessarily since the Chrome path normally doesn't change during a process runtime. Caching the result improves initialization speed and minimizes disk I/O operations.

Measured Improvement:
A benchmark looping 10,000 times showed:

Before (uncached): ~30.3ms
After (cached): ~1.4ms
This represents a >20x speedup in the target code path.

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,0 +1,3 @@
+## 2023-10-27 - [Cache resolved system Chrome path]
+**Learning:** `existsSync` checks within `packages/shared/src/mcp/chrome-path.ts` were repeatedly called when resolving the Chrome path.
+**Action:** Introduced a module-level variable to cache the result of the first successful resolution to avoid repeated synchronous filesystem hits in the same process.
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,3 +1,0 @@
-## 2023-10-27 - [Cache resolved system Chrome path]
-**Learning:** `existsSync` checks within `packages/shared/src/mcp/chrome-path.ts` were repeatedly called when resolving the Chrome path.
-**Action:** Introduced a module-level variable to cache the result of the first successful resolution to avoid repeated synchronous filesystem hits in the same process.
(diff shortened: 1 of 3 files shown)


quanru:
Thanks for the contribution! I took a close look and I have some concerns about whether this change is a net positive.

**The benchmark doesn't match any real call pattern.**
I grepped the repo and `getSystemChromePath()` (via its wrapper `resolveChromePath()`) is only called from a single place: `launchDetachedChrome()` in `packages/web-integration/src/mcp-tools-puppeteer.ts`. That means it runs at most **once per Chrome launch**, not in a loop. The 10k-iteration benchmark in the PR description doesn't correspond to any actual code path — there's no hot path to optimize here. The function itself does up to 7 `existsSync` checks, which is in the microsecond range for a single call.

**Cost side of the change:**
1. It introduces module-level mutable state for a function that runs once per process lifetime in practice.
2. It exports a new public API (`clearSystemChromePathCache`) that only exists to keep tests isolated. That's a test smell — production code shouldn't grow public surface to serve test setup. `vi.resetModules()` or a closure-based container would handle this without a new export.
3. The cache only stores successful resolutions — if Chrome isn't found, every call re-scans the full path list. This semantic isn't stated or tested. Is that intentional?
4. No test actually verifies the caching behavior itself (e.g., that a second call doesn't re-invoke `existsSync`, or what happens on platform change). The new `clearSystemChromePathCache` call in `beforeEach` only prevents test pollution; it doesn't prove the cache works.

**Suggestion:**
I'd lean toward not merging this unless there's a concrete user-facing scenario where `getSystemChromePath` is called repeatedly. If there is one, it would be better to:
- Call the expensive function once at the caller site and reuse the result, rather than caching inside the utility; **or**
- Keep the cache but define and test the "not found" semantics, and avoid exporting a test-only reset helper.

Happy to reconsider if I'm missing a call site where this matters in practice.

quanru:
Closing this one — as noted in the review above, `getSystemChromePath` is only called once per Chrome launch, so there's no real hot path to optimize, and the module-level cache plus test-only exported reset helper adds complexity without a corresponding benefit. Thanks again for taking the time to send a patch; we'd welcome future contributions that target measured bottlenecks in real call paths.

PR #2338, opened 2026-04-10
perf:optimize-waterflow-dom-query

vincerevu:
**What:** Added caching logic for `this.pointerElement` in `water-flow.ts` to reuse the existing element without performing expensive DOM queries repeatedly.

 **Why:** The water flow animation creates an overlay pointer that moves based on user interactions. Constantly querying `document.querySelector` on every cycle added unnecessary CPU overhead. By caching the element and protecting against detached nodes, we make the animation more efficient.

 **Measured Improvement:** We ran a benchmark loop simulating 1,000,000 toggles.
- **Baseline time:** ~88.82 ms (1,500,000 querySelector calls)
- **Optimized time:** ~56.90 ms (500,000 querySelector calls)
- **Improvement:** Reduced DOM queries by 66% and overall execution time by ~36%.


quanru (commented):
Thanks for the contribution! I took a close look at both the diff and the surrounding `water-flow.ts`, and I have some concerns about whether the caching change is worth merging as-is.

## The measured gain doesn't translate to real usage

The benchmark shows 88ms → 57ms over **1,000,000 iterations**, which works out to roughly **30 nanoseconds saved per call**.

In actual usage, `showMousePointer` / `hideMousePointer` are invoked while Midscene drives the cursor — a few times per second at most, not a million times. At that frequency, the savings are far below anything a user or the browser can perceive. The "36% improvement" is mathematically correct but describes a denominator that doesn't exist in the real hot path. `document.querySelector('div[attr]')` on `<body>` is already in the sub-microsecond range; it's not a bottleneck here.

Our repo guideline is explicit about this:
> Don't add features, refactor, or introduce abstractions beyond what the task requires.

## The cache adds a consistency burden that outweighs the (non-)gain

Introducing `pointerElement` means we now have to keep a module-level reference in sync with the DOM in three places:
- `showMousePointer` (detached-node check via `document.body.contains`)
- the fade-out `setTimeout` (null out after `removeChild`)
- `disable()` (null out after cleanup)

Every new cache is a new invariant that future maintainers must preserve on every exit path. Paying that long-term cognitive cost to save ~30ns per call isn't a good trade.

## There's also a small gap in the cache logic itself

In `showMousePointer`:

```ts
const pointer = existingPointer || (() => { /* ... */ this.pointerElement = p; return p; })();
```

`this.pointerElement` is only written in the **creation** branch. If `existingPointer` came from the `querySelector` fallback (i.e. the cache was empty), the found node is **not** written back into the cache, so the next call still has to fall back to `querySelector`. Not a bug, just evidence that the caching layer isn't fully load-bearing.

## What is worth keeping

These two changes are genuine robustness fixes, independent of the caching optimization:

- `hideMousePointer`: guarding `removeChild` with `pointer.parentNode`
- `disable()`: guarding `removeChild` with `element.parentNode`

These protect against a real race: the fade-out `setTimeout` removing the node before `hideMousePointer` / `disable` runs, which would otherwise throw `NotFoundError`. Happy to merge these right away.

## Suggested next step

Could you split this into a smaller PR that keeps only the `parentNode` safety checks and drops the `pointerElement` caching? If you still believe the caching is worthwhile, please share a profile from a real Midscene run (not a synthetic 1M-iteration loop) showing that `querySelector` shows up meaningfully on the hot path — I suspect it won't.

One more nit: the commit title `perf:optimize-waterflow-dom-query` doesn't satisfy our Conventional Commits scope requirement. Per `commitlint.config.js`, this should be something like `perf(chrome-extension): ...`.

Thanks again for looking into this!

PR #2348, opened 2026-04-14
chore(android): use local proxy when downloading binary files

### LINKED ISSUE #2347: [Feature]: Detecting the user proxy and reuse in the installation script
### What problem does this feature solve?

In some network environments, github.com cannot be accessed directly. This can cause the downloading of binary files via download-scrcpy-server.mjs and download-yadb.mjs to fail. 

We could consider detecting the user proxy during installation and then using the proxy for downloading.

### What does the proposed API look like?

No API modifications, only enhancements to the installation script.

orozot:
# Summary
Add proxy support to the Android prebuild scripts for downloading `scrcpy` and `yadb` binaries. Fix #2347 

# Key Changes
1. Introduced a new dependency: `https-proxy-agent@7.0.6`
2. Updated `packages/android/scripts/download-scrcpy-server.mjs`:
    - Detect proxy configuration from environment variables
    - Apply proxy settings to `fetchVersion` when a proxy is present
3. Updated `packages/android/scripts/download-yadb.mjs`:
    - Detect proxy configuration from environment variables
    - Apply proxy settings to `fetchVersion` when a proxy is present
 
# Testing
- [ ] Set `HTTPS_PROXY` and `HTTP_PROXY`, then run:
```bash
node packages/android/scripts/download-yadb.mjs
node packages/android/scripts/download-scrcpy-server.mjs
```
- [ ] Unset `HTTPS_PROXY` and `HTTP_PROXY`, then run:
```bash
node packages/android/scripts/download-yadb.mjs
node packages/android/scripts/download-scrcpy-server.mjs
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,6 @@ packages/all-in-one-playground/static/*
 # rslib
 .rslib
 **/.rslib
+
+# jetbrains
+.idea/
--- a/packages/android/package.json
+++ b/packages/android/package.json
@@ -64,7 +64,8 @@
     "typescript": "^5.8.3",
     "tsx": "^4.19.2",
     "vitest": "3.0.5",
-    "zod": "^3.25.1"
+    "zod": "^3.25.1",
+    "https-proxy-agent": "7.0.6"
   },
   "license": "MIT"
 }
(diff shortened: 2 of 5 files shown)


yuyutaotao:
Hi @orozot 

Thanks for the contribution.

We reworked this part on our side in https://github.com/web-infra-dev/midscene/pull/2363 , mainly to align the download scripts with the same dependency approach we already use in core, and to use the more common proxy environment variable names such as http_proxy / https_proxy (and their uppercase variants).

PR #2381, opened 2026-04-21
fix(web-integration): derive Playwright reportFileName from readable test title

quanru:
## Summary

- `PlaywrightAiFixture` was forwarding Playwright's internal `testInfo.testId` (a UUID) as the Agent's `reportFileName`, producing intermediate report files named `playwright-<uuid>-<uuid>.html` that carry no information about which spec or case produced them — impossible to triage on disk, and it breaks downstream parsing scripts that expect a readable `playwright-<title>-<date>-<shortUuid>.html` pattern.
- Route `reportFileName` through the already path-safe, retry-aware `title` returned by `groupAndCaseForTest`. Keep the Agent `testId` option in lockstep with `reportFileName` so the deprecated `testId` fallback in `@midscene/core` cannot silently reintroduce a UUID-based filename if `reportFileName` is ever dropped.
- Reproduced via the new unit test file. Before: `reportFileName = playwright-<playwright-testId-uuid>-<page-uuid>`. After: `reportFileName = playwright-<readable-title-with-retry>-<page-uuid>`.

## Root cause

`packages/web-integration/src/playwright/ai-fixture.ts` destructured `testInfo.testId` (Playwright's internal UUID) and used it to build the Agent's `reportFileName`, ignoring the human-readable `title` / `id` fields that `groupAndCaseForTest` was already computing right beside it. The Agent then wrote that UUID directly to disk because `ReportGenerator.create()` treats an explicit `reportFileName` as the final filename (no date/shortUuid wrapping).

Behavior regressed in two steps: `#2264` switched the field from `testId` to `reportFileName` (now used verbatim), and `#2328` kept both fields carrying the UUID.

## Test plan

- [x] New repro/regression file `packages/web-integration/tests/unit-test/playwright-ai-fixture-report-filename.test.ts` (4 cases):
  - Readable title replaces UUID `testId`; `testId` option stays in lockstep with `reportFileName`.
  - Retry marker `(retry--N)` is embedded so reruns do not clobber the first attempt.
  - Multi-page agents in one test do not collide (page-level uuid suffix preserved).
  - Filename-hostile characters (`:*?"<>|#` and space) are scrubbed; path separators `/\` are preserved (matches `replaceIllegalPathCharsAndSpace`).
- [x] `npx vitest run packages/web-integration/tests/unit-test/playwright-ai-fixture-*.test.ts` — 21/21 passing.
- [x] `pnpm run lint` — clean (only one pre-existing unrelated warning in `packages/core`).

### DIFF
--- a/packages/web-integration/src/playwright/ai-fixture.ts
+++ b/packages/web-integration/src/playwright/ai-fixture.ts
@@ -137,7 +137,11 @@ export const PlaywrightAiFixture = (options?: PlaywrightAiFixtureOptions) => {
       (page as any)[midsceneAgentKeyId] = idForPage;
       const { file, title } = groupAndCaseForTest(testInfo);
       const cacheConfig = processTestCacheConfig(testInfo);
-      const reportTag = `playwright-${title}-${idForPage}`;
+      // `replaceIllegalPathCharsAndSpace` intentionally preserves `/` and `\`
+      // so groupName/groupDescription can still carry hierarchy. But
+      // ReportGenerator rejects path separators in the file name, so strip
+      // them here for the report tag only.
+      const reportTag = `playwright-${title.replace(/[\\/]/g, '-')}-${idForPage}`;
 
       const agent = new PlaywrightAgent(page, {
         testId: reportTag,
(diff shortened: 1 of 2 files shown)


quanru:
Addressed the Codex review feedback in 80d3646:

- `replaceIllegalPathCharsAndSpace` deliberately preserves `/` and `\`, but `ReportGenerator.validateReportFileName` rejects both. A test title like `"login / signup"` would therefore crash fixture init with `reportFileName must not contain path separators`.
- Strip `/` and `\` only when building `reportTag`; `groupName`/`groupDescription` still keep the original separators so hierarchy displayed in reports is unaffected.
- Added a regression case that feeds a title containing both `/` and `\`, asserts the resulting `reportFileName` is separator-free while the surrounding words survive, and confirms `groupName` still carries the raw separators.

yuyutaotao (approved):

PR #2418, opened 2026-04-28
test(studio): add startup smoke coverage

quanru:
## Summary
- add packaged startup smoke coverage for `apps/studio`
- add headless Linux CI wiring for the Studio startup path
- add a Midscene-driven Studio startup smoke that uploads `studio-startup-ai-report`
- align Studio tests with the rebased device discovery and preload bridge APIs

## Validation
- `node --check apps/studio/scripts/startup-smoke-ai.mjs`
- `NX_DAEMON=false npx nx test studio`
- `npx nx build studio`
- `MIDSCENE_STUDIO_RUN_STARTUP_SMOKE=1 pnpm --dir apps/studio run test:smoke`
- `pnpm run lint`
- manual GitHub Actions dispatch: `Headless Linux` with `suite=studio` on `codex/studio-headless-smoke` (`25046917796`), passed with `studio-startup-ai-report` artifact
- PR checks on `2c944a8ad`: passed

PR #2448, opened 2026-05-08
fix(studio): enable manual preview control for Computer

quanru:
## Summary

- Computer was the only Studio platform left out of the manual mouse/touch matrix added in #2433 (Android/iOS/Harmony) and #2442 (Web). Clicks on the middle preview area for a connected Computer session were silently dropped.
- The underlying plumbing was already in place: `@midscene/computer`'s `actionSpace` exposes `Tap`, `KeyboardPress`, `Input` and `DragAndDrop`, and `agent.interface.size()` returns the display dimensions through `/interface-info`. The gap was purely a renderer-side platform allowlist in `MainContent`.
- Extract the three platform predicates (`isManualPreviewControlSupported`, `resolveManualDragActionType`, `isManualPreviewKeyboardSupported`) into `preview-layout.ts` so the matrix lives in one place, and call them from `MainContent`.
- Computer routes drag through `DragAndDrop` (it does not expose `Swipe`, unlike mobile platforms — see `packages/computer/src/device.ts:752`) and opts into the keyboard sink alongside Web.

## Test plan

- [x] `npx vitest run tests/main-content-preview-layout.test.ts` (17 passed; 9 new cases covering the three predicates across `android`/`ios`/`harmony`/`computer`/`web`/`undefined`)
- [x] `npx vitest run tests/main-content-overview.test.tsx tests/main-content-preview-layout.test.ts tests/main-content-web-navigation.test.tsx` (24 passed)
- [x] `npx tsc --noEmit` in `apps/studio`
- [x] `pnpm run lint` (clean — only a pre-existing warning in `packages/core/tests/unit-test/report-generator-async-contract.test.ts`)
- [ ] Manual: connect a Computer session in Studio, tap / drag / type on the preview, confirm `/interact` calls land. (Requires local accessibility permission; not exercised in CI.)

### DIFF
--- a/apps/studio/scripts/startup-smoke-ai.mjs
+++ b/apps/studio/scripts/startup-smoke-ai.mjs
@@ -228,7 +228,7 @@ async function main() {
     );
 
     await agent.aiAssert(
-      'The right-side Playground setup panel is fully visible and is not clipped by the screenshot edge. The welcome title, platform cards, ADB device field, and Create Agent button should all be readable without being cut off on the right.',
+      'The right-side Playground panel is fully visible and is not clipped by the screenshot edge. The welcome empty state with the heading "Welcome to Midscene.js Playground!" should be readable without being cut off on the right.',
     );
 
     await agent.recordToReport('Studio shell became visible', {
@@ -242,7 +242,7 @@ async function main() {
     await sleep(1_500);
 
     await agent.aiAssert(
-      'A settings popup panel is visible and contains the items Language, Theme, GitHub, Website, and Environment.',
+      'A settings popup panel is visible and contains the items Language, Theme, GitHub, and Website.',
     );
 
     await agent.destroy();
--- a/apps/studio/src/renderer/components/MainContent/index.tsx
+++ b/apps/studio/src/renderer/components/MainContent/index.tsx
@@ -29,9 +29,6 @@ import type { ShellActiveView } from '../ShellLayout/types';
 import { DeviceList } from './DeviceList';
 import { MobilePreviewFrame } from './MobilePreviewFrame';
 import {
-  isManualPreviewControlSupported,
-  isManualPreviewKeyboardSupported,
-  resolveManualDragActionType,
   resolveStudioPreviewPlatform,
   shouldEnableMobilePreviewFrame,
   shouldUseDesktopPreviewPadding,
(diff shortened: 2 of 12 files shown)


quanru:
Pushed two updates after looking at the failing CI runs:

1. **Rebased on latest `main`** so the new commits from #2449 are picked up. Re-ran `npx tsc --noEmit` and the `apps/studio/tests/main-content-*` suites locally, all green.
2. **Updated the AI smoke prompts in `apps/studio/scripts/startup-smoke-ai.mjs`** to match the new shell UI. The earlier prompts from #2418 still asserted on welcome-screen elements (`ADB device field`, `Create Agent button`) and a settings popup item (`Environment`) that #2437/#2449 either moved into a per-device setup form or relocated to the dedicated `Env` row in `SettingsDock` — so the smoke had been failing on every PR since #2449 merged (e.g. https://github.com/web-infra-dev/midscene/actions/runs/25589106473 on `feat/studio-i18n` shows the same `expected 1 to be +0` at the same line). Strictly speaking this is post-#2449 fallout rather than this PR's territory, but bundling it here unblocks merging without a separate PR.

The other failing run (`AI unit test` → `tests/ai/web/puppeteer/ai-number.test.ts`) looks like an AI flake — same test passed on this branch's earlier run (25544333063) and on `feat/studio-i18n` (25589106442) at the same time window. Re-running that workflow should be enough.

quanru:
Pushed a refactor commit (`ee3c8aebd`) that gets rid of the platform-keyed table I introduced in the first pass.

**Why**: the table (`MANUAL_CONTROL_PLATFORMS` / drag flavor / keyboard) was a renderer-side mirror of each device's actionSpace, which the server already knows. New platform → three places to update; future actionSpace changes → renderer drift.

**What changed**:

- `/interface-info` now reports the device actionSpace as `actionTypes: string[]` (in `packages/playground/src/server.ts`); SDK + both adapters carry the field through.
- `PreviewRenderer` pulls `actionTypes` alongside the device `size` it was already fetching, and self-derives:
  - `manualControl = actionTypes.includes('Tap')`
  - `drag flavor = 'Swipe' if exposed else 'DragAndDrop'` — matches today's behaviour without a platform map (mobile actionSpaces ship both, Computer / Web ship only `DragAndDrop`)
  - `keyboard = KeyboardPress or Input present`
- `PlaygroundPreview` drops the `manualControlEnabled` / `manualDragActionType` / `manualKeyboardEnabled` props.
- Studio drops the `MainContent` derivation, the three `preview-layout` helpers, and their unit tests.
- Coverage: a new `/interface-info` test asserts `actionTypes` is exposed; the renderer manual-input test now drives behaviour through mocked `actionTypes` instead of props.

**Side effect to flag**: standalone `@midscene/playground-app` consumers (non-Studio) now get manual preview control automatically when their device exposes `Tap`. The previous 'off by default' came from those callers never opting in via the props that no longer exist. I think this is the right behaviour — the underlying server already accepts `/interact` for any actionSpace member — but it is a behaviour change for non-Studio users, hence calling it out.

Validation: `npx nx test @midscene/playground` (212 ✓), `npx nx test @midscene/playground-app` (48 ✓), `apps/studio` `tsc --noEmit` clean, `pnpm run lint` clean, Studio `main-content-*` suites green.

yuyutaotao (approved):

PR #2488, opened 2026-05-18
feat(core): support MIDSCENE_CUSTOM_SYSTEM_PROMPT env var for custom system prompt injection

will-:
## Summary

- Add `MIDSCENE_CUSTOM_SYSTEM_PROMPT` environment variable support that prepends a custom system prompt to all AI calls (planning, locate, extract, assert)
- Enable private/self-hosted models to receive model-specific system instructions without code changes
- The custom prompt is injected at the `callAI()` entry point, covering all downstream paths including CodexAppServer

## Motivation

When using private or self-hosted models, some require specific system prompt instructions (e.g., role definitions, output format constraints) that differ from Midscene's built-in prompts. Currently there is no way to inject custom system-level instructions without modifying source code.

This change allows users to simply set an environment variable:

```bash
export MIDSCENE_CUSTOM_SYSTEM_PROMPT="You are a UI automation assistant. Always respond in valid JSON."
```

## Changes

- `packages/shared/src/env/types.ts` — Declare `MIDSCENE_CUSTOM_SYSTEM_PROMPT` constant and register in `STRING_ENV_KEYS`
- `packages/core/src/ai-model/service-caller/index.ts` — Add `applyCustomSystemPrompt()` function and call it at `callAI()` entry point
- `packages/core/tests/unit-test/custom-system-prompt.test.ts` — 8 unit tests for the pure function
- `packages/core/tests/unit-test/custom-system-prompt-integration.test.ts` — 4 integration tests mocking OpenAI client to verify end-to-end behavior

## Test plan

- [x] `pnpm run lint` — passes (no new lint errors)
- [x] Unit tests: `vitest --run tests/unit-test/custom-system-prompt.test.ts` — 8 passed
- [x] Integration tests: `vitest --run tests/unit-test/custom-system-prompt-integration.test.ts` — 4 passed
- [x] Full core test suite: 67 test files, 846 tests passed, 0 failures


EAGzzyCSL:
Could you share the concrete case this feature is meant to solve?

If the main goal is to constrain the response format, for example requiring valid JSON, Midscene’s existing prompts already include output format requirements, so in theory the model should be able to follow them. If a model still cannot return valid JSON under the existing format constraints, my bigger concern is that this may not be a single prompt issue. Even if an extra system prompt fixes the JSON format, the same model may still run into stability issues in task understanding, element locating, structured extraction, coordinate/action generation, and other parts of the workflow.

So I’d like to first clarify whether this is for a more specific model-adaptation case, such as a private model that requires certain system instructions to work properly, or whether the main intention is to strengthen output format constraints.

quanru:
Thanks for the contribution.

After evaluating this internally, we have decided to close this PR for now.

The main reason is that the current motivation is still too broad. If the goal is to make model responses follow a stricter format, such as always returning valid JSON, Midscene already has output format requirements in the built-in prompts and parsing/retry logic. In that case, adding a global extra system prompt may hide a deeper model compatibility issue rather than solve it reliably.

This feature may still make sense for a concrete model-adaptation case, for example a private or self-hosted model that requires specific system-level instructions before it can work properly with Midscene. If you can share the exact model/provider behavior, the failure mode, and the prompt or request requirement that makes this necessary, we can re-evaluate the design based on that scenario.

For now, we will close this PR. Feel free to open a new PR or continue the discussion with a concrete use case.

PR #2513, opened 2026-05-21
fix(core): throw on out-of-range bbox from normalized-[0,1000] models

quanru:
## Summary

- When `MIDSCENE_MODEL_FAMILY` routes through `normalized01000` in `adaptBbox` (qwen3-vl / qwen3.5 / qwen3.6 / glm-v / auto-glm{,-multilingual}), enforce that each bbox component sits in `[0, 1000]` and throw with `(modelFamily, raw bbox, shotSize)` otherwise.
- The guard only fires when `modelFamily` is explicitly provided. Existing `adaptBboxToRect` boundary-clamp behaviour (used in unit tests, called without a family) is preserved.
- Added 8 new tests in `packages/core/tests/unit-test/adapt_bbox.test.ts` covering in-range, boundary, out-of-range, negative, NaN, non-array, and the qwen2.5-vl negative case.

## Why

Observed in a Mi Car app yaml run (Midscene CLI 1.8.3, `tongyi/qwen3.6-plus`): the model occasionally emitted pixel-style coords like `[0, 500, 1080, 1920]` for the \"scrollable list region\". `adaptBbox` happily fed those into `normalized01000` against a 720×1600 shot, producing `center = [389, 1936]` (well outside the screenshot). The resulting ADB `input swipe 778 4268 778 2167` landed off-screen, the page never scrolled, the agent kept replanning, and the run ended with a generic \"Replanned 20 times, exceeding the limit\" error — the real signal (\"model returned out-of-range bbox\") was lost.

The change matches the project design principle in `AGENTS.md`: *Throw errors instead of returning blank values when something goes wrong.*

## Test plan

- [x] `cd packages/core && pnpm exec vitest run tests/unit-test/adapt_bbox.test.ts` — 37 tests pass (8 new).
- [x] `pnpm run lint` (root) — clean.
- [ ] Optional: an integration test against a live qwen3.6 endpoint to confirm the new error surfaces in `aiAct` task output (skipped here — disk space / model access).

## Notes for reviewers

- Error message includes a remedy hint (\"switch MIDSCENE_MODEL_FAMILY to qwen3-vl, or revisit the bbox prompt\") so users can self-triage.
- We intentionally do NOT auto-detect / rescale pixel coords. Self-healing here would hide model regressions; the explicit failure is the desired UX.
- Same pattern can be extended later to other families' bbox validators if similar regressions appear.

### DIFF
--- a/packages/core/src/common.ts
+++ b/packages/core/src/common.ts
@@ -235,11 +235,49 @@ export function adaptBbox(
     result = adaptGpt5Bbox(normalizedBbox);
   } else {
     // Default: normalized 0-1000 coordinate system
-    // Includes: qwen3-vl, qwen3.5, glm-v, auto-glm, auto-glm-multilingual, and future models
+    // Includes: qwen3-vl, qwen3.5, qwen3.6, glm-v, auto-glm, auto-glm-multilingual,
+    // and future models.
+    if (modelFamily !== undefined) {
+      assertNormalized01000Bbox(normalizedBbox, modelFamily, width, height);
+    }
     result = normalized01000(normalizedBbox as number[], width, height);
   }
 
   return result;
+}
+
+// Validate a bbox that is expected to be in the normalized [0, 1000] coordinate
+// system. Some grounding models (notably qwen3.5 / qwen3.6) occasionally emit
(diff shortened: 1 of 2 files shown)


quanru:
Reopening from upstream branch so CI can access secrets. Continued in a new PR.

PR #2518, opened 2026-05-22
fix(web-integration): prevent dropped characters when input re-renders mid-type

quanru:
## Summary

`typeText` in **replace** mode (the default for the `Input` action) calls `clearInput` and then immediately runs `keyboard.type`. `clearInput` fires a synthetic `input` event with `value === ""`, which many controlled components react to by re-rendering and sometimes replacing the input element entirely. If that replacement lands after `clearInput` returns but during `keyboard.type`, characters can land on the detached old element and be dropped.

## Fix

- Add `Page.waitForDomQuiet({ quietMs, timeoutMs, target })` to `packages/web-integration/src/puppeteer/base-page.ts`. It uses a `MutationObserver` scoped to the target input's stable ancestor (`closest("form")`, then parent element, then `document.body` as fallback), so unrelated page-level mutations do not keep resetting the quiet timer.
- Declare `waitForDomQuiet` as an optional method on `AbstractWebPage` so shared input code can call it safely.
- In `createWebInputPrimitives` replace mode, call `await page.waitForDomQuiet?.({ target })` between `clearInput` and `keyboard.type`. Async re-renders triggered by clearing are now absorbed before typing starts.

## Regression test

`packages/web-integration/tests/unit-test/puppeteer/input-clear-then-type-race.test.ts` drives a real Puppeteer browser against a page that replaces the input element after the clear-triggered `input` event. The production `Input` action path now preserves the complete value (`Hello`).

The same test file also covers the scoped observer behavior: unrelated mutations outside the target ancestor no longer force the wait to run until the hard timeout.

## Test plan

- [x] `pnpm run lint`
- [x] `npx nx test @midscene/web -- input-clear-then-type-race.test.ts`


ottomao (changes requested):
Could we just use a hard sleep ?

quanru:
@ottomao Thanks, I checked the hard-sleep option.

I do not think a fixed 300ms sleep is sufficient for this race. In the reproduced case, the input is replaced 450ms after the clear-triggered `input` event; replacing the DOM-quiet wait with a 300ms sleep still loses characters (`Hello` becomes `ello`). A longer fixed sleep would be a guess and would add unconditional latency to every replace-mode input.

I kept the DOM-quiet wait, but narrowed its scope in `15a1239e2`: it now observes the target input’s stable ancestor (`closest("form")`, then parent element, with `document.body` only as fallback) instead of the whole page. This preserves coverage for `parent.replaceChild(input)` while avoiding unrelated page-level mutations from resetting the timer.

I also added coverage for the scoped behavior in `input-clear-then-type-race.test.ts`.

Validation:
- `npx nx test @midscene/web -- input-clear-then-type-race.test.ts`
- `pnpm run lint`

ottomao (approved):

PR #2525, opened 2026-05-22
fix(site): tune Bilibili and iQIYI logo widths

quanru:
## Summary
- The Latin and Chinese Bilibili wordmarks have very different aspect ratios (about 2.24:1 vs 3.53:1), so a single 140px width made the Chinese logo render visually heavier than the surrounding icon/text entries. Drive `logoWidth` through the existing i18n map (`userBilibiliLogoWidth`): 90px for the Latin wordmark and 120px for the Chinese wordmark, which lines each version up with the rest of the row.
- The iQIYI wordmark has an unusually wide aspect ratio (about 6.47:1), so the previous 140px width rendered it at only ~22px tall — visibly smaller than the surrounding icons and text. Bump the width to 200px so it sits closer to ~31px tall, in line with other entries.

## Test plan
- [x] `pnpm run lint`
- [ ] Visual check of the homepage in both `en` and `zh` locales (please verify in preview)

### DIFF
--- a/apps/site/theme/components/WhoIsUsing.tsx
+++ b/apps/site/theme/components/WhoIsUsing.tsx
@@ -110,7 +110,7 @@ export function WhoIsUsing() {
       name: t('userIqiyi'),
       url: 'https://www.iqiyi.com',
       logo: '/images/users/iqiyi-color.svg',
-      logoWidth: 140,
+      logoWidth: 200,
     },
     {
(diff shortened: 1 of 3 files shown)


quanru:
Reopening as two separate PRs to keep Bilibili and iQIYI logo size changes independent.

EAGzzyCSL (approved):

PR #2535, opened 2026-05-25
fix(playground): propagate AbortSignal through cancel chain

quanru:
## Summary

After running multiple steps in the Computer playground and clicking Stop, the
mouse/keyboard would keep being driven — because `/cancel` only ran
`agent.destroy()` + recreate and never told the in-flight `executeAction →
callActionInActionSpace → device action` to stop. The fix turns the cancel
path into a cooperative abort chain.

## Commit order (intentional)

1. **`test(computer): add regression coverage ...`** — adds the e2e test, the
   focused unit test, the `headless-linux-cancel` CI job, and wires
   `@midscene/playground` as a devDependency of `packages/computer`. **No fix
   yet** — this commit is meant to turn the new `headless-linux-cancel` job
   red on CI, proving the test catches the regression.
2. **`test(computer): set dummy model env ...`** — small follow-up: the test
   never calls a model, but `PlaygroundServer` validates
   `MIDSCENE_MODEL_NAME` at startup, so dummy values are needed for the
   headless runner to even reach the abort assertion.
3. **`fix(playground): ...`** (next push) — the actual L1–L5 fix:
   - L1 server: `taskAbortControllers` map + `/cancel` calls
     `controller.abort()` before falling back to destroy+recreate
   - L2 `executeAction`: forwards `abortSignal` to all three branches
   - L3 core: `callActionInActionSpace`, `runPlans`, and `Agent.destroy`
     plumb / fire the signal
   - L4 `ComputerDevice`: `destroy()` sets `destroyed=true` first; internal
     `sleep` loops in `smoothMoveMouse / performScroll / typeText /
     dragAndDrop` become abortable
   - L5 visualizer `handleStop`: flips `interruptedFlagRef` and
     `setLoading(false)` synchronously before awaiting the cancel POST

Once the fix lands, `headless-linux-cancel` should turn green.

## Test plan

- [ ] CI: `headless-linux-cancel` red on commits 1+2 (regression caught)
- [ ] CI: `headless-linux-cancel` green on commit 3 (fix verified)
- [x] Local mac: three-phase verification done (baseline pass → comment out
      fix to reproduce → restore fix to verify)
- [x] `pnpm run lint` clean before each commit


yuyutaotao:
Try this instead https://github.com/web-infra-dev/midscene/pull/2548

PR #2582, opened 2026-06-02
feat(testing-framework): AI-native UI testing framework on Rstest

quanru:
## What

Adds **`@midscene/testing-framework`** — a thin layer over Rstest that turns a `midscene.config.ts` plus natural-language YAML cases into a test run — and wires it into the `midscene` CLI. Three parts (folded in from the former #2583 / #2584 stack):

1. **`@midscene/testing-framework`** — programmatic `runMidsceneTest` (run a config's YAML cases through Rstest in-process) and `emitRstestProject` (export a self-contained native Rstest project: `rstest.config.ts` + `e2e/*.test.ts` + `package.json`).
2. **`midscene emit <out-dir> [--config <path>]`** — CLI command on top of `emitRstestProject`.
3. **`@midscene/shared/rstest`** — the in-process Rstest runner (`buildRstestInlineConfig` + `runRstestWithVirtualModules`) shared by both `@midscene/cli`'s YAML path and `@midscene/testing-framework`, so the virtual-module + `VirtualModulesPlugin` wiring lives in one place.

Dependency direction stays acyclic: **`cli → testing-framework → shared`**. `testing-framework` never depends on `@midscene/cli`.

## Why this design

The user's `midscene.config.ts` often imports Playwright, whose bundled `@vitest/expect` non-configurably registers `Symbol.for('$$jest-matchers-object')` on `globalThis`. Loading the config in the **same process** that then calls `runRstest` makes Rstest's own `@vitest/expect` crash on redefining that symbol — the reason an earlier approach forked a child process.

**Key insight:** the runner only resolves the config *path*; the config is imported **only inside the Rstest worker** via a generated bootstrap module (top-level `await` discovers cases and registers one `test()` each against a shared agent). The main process stays clean, so `runRstest` runs **in-process with no fork**. The runtime never statically imports `@rstest/core` — the test API is passed in as a parameter — keeping its expect global out of the main process too.

## Notable details

- The Rstest worker owns results: it writes a single `summary.json` the runner reads (no per-case re-aggregation).
- `testRunner.testTimeout` / `retry` are forwarded per case; `bail` is enforced case-by-case; `setup`/`beforeAll` failures surface via `summary.suiteError`.
- **`cache: true` convenience:** Core rejects `cache: true` (it never invents ids). The framework derives a stable id from the project folder name in `setupFrameworkAgent`, so users keep writing `cache: true` and get a reusable cache; an explicit `cache: { id }` is left untouched.
- Default `web` (PlaywrightAgent) / `android` (`agentFromAdbDevice`) targets or a custom `setup`; custom `yamlSteps` interleave with built-in steps; `.env` auto-loading.
- Shared runner stays free of a hard `@rstest/core` / `@rsbuild/core` dependency: each caller resolves them from its own base (the user project for `testing-framework`, the bundled copy for `cli`) and passes the `@rsbuild/core` entry path in. `@rstest/core` is dev-only (types), pinned to `0.10.3` everywhere.

## Validation

- Builds: `npx nx build shared testing-framework cli` all green; `node scripts/check-tsconfig-references.mjs` clean (cli references testing-framework).
- Unit: `shared` 355, `testing-framework` 58, `cli` framework + emit 37 — all pass. `pnpm run lint` clean.
- End-to-end against a prepatch beta on `testing-framework-demo` with **real AI**, all demos back to `cache: true` (framework derives the stable id): see PR comment for the run.

Draft — opening for review before finalizing.


quanru:
## End-to-end verification (prepatch `1.8.8-beta-20260603024013.0`)

Verified the full `midscene emit` → `rstest run` flow on `midscene-example` testing-framework demos with **real AI (Doubao vision)**, all demos using `cache: true` (framework derives the stable id):

| Demo | Flow | Result |
|---|---|---|
| framework-with-setup (web, custom setup importing Playwright) | `midscene emit` → `rstest run` | **2/2 pass**, 84.5s cold |
| — second run | cache hit | **2/2 pass**, 27.7s (3× faster, `cache hit` logged) |
| custom-yaml-steps (web, target + custom yamlSteps) | same | **1/1 pass**, 23.2s |
| android-config-demo | `midscene emit` | product correct, `testTimeout: 300000` derived (full run needs a device) |

- `emit` produces `rstest.config.ts` + `midscene.config.ts` + `e2e/*.test.ts` + `package.json` (`test: rstest run`), version pinned to the beta.
- `cache: true` is accepted (Core would otherwise throw): cache written to `midscene_run/cache/<dir>.cache.yaml` with `cacheId` derived from the project folder name.
- Demo migration (drop the removed `midscene-testing-framework` bin → `midscene emit` + `rstest run`) lives in midscene-example PR #96.

PR #2584, opened 2026-06-02
refactor(shared): share the in-process Rstest runner between cli + testing-framework

quanru:
Stacked on #2582 (base = `codex/testing-framework-native-rstest`).

## What
`@midscene/cli` (`runRstestYamlProject`) and `@midscene/testing-framework` (`defaultBootstrapRunner`) had the same ~15-line kernel: build an Rstest `inlineConfig` that registers virtual test modules via `@rsbuild/core`'s `VirtualModulesPlugin`, then call `runRstest`. This extracts that kernel to **`@midscene/shared/rstest`** and points both at it.

- `buildRstestInlineConfig(options, rsbuild)` — pure, unit-tested config builder.
- `runRstestWithVirtualModules(options)` — resolves `@rstest/core/api` at runtime + calls `runRstest`.

## Why it's safe (no cycle, no new runtime dep)
- `@midscene/shared` is already a dependency of both packages, so there's no cycle.
- Each caller keeps its **own** resolution strategy and passes the resolved `@rsbuild/core` entry in: the framework resolves from the **user project** (`@rstest/core` peer dep), the CLI from its **bundled** copy. So `@midscene/shared` only needs `@rstest/core` as a **dev-only** dependency (types) — it never ships a runtime dependency on it; at runtime the module resolves from the caller's context.

## Validation
- `nx build shared|testing-framework|cli` all pass; `check-dependency-version` clean; `pnpm lint` clean.
- New `packages/shared/tests/unit-test/rstest.test.ts` (3 tests) for the pure builder.
- `testing-framework` 58 unit tests pass; all `cli` framework tests pass — including the `reporters: []` assertion and the real `runRstest` virtual-module runs now routed through the shared helper.

Draft — review after #2582.

### DIFF
--- a/packages/cli/src/framework/rstest-runner.ts
+++ b/packages/cli/src/framework/rstest-runner.ts
@@ -1,7 +1,7 @@
 import { createRequire } from 'node:module';
 import { dirname, join, resolve } from 'node:path';
-import { pathToFileURL } from 'node:url';
-import type { RstestUserConfig, TestRunResult } from '@rstest/core/api';
+import { runRstestWithVirtualModules } from '@midscene/shared/rstest';
+import type { TestRunResult } from '@rstest/core/api';
 import type { GeneratedRstestYamlProject } from './rstest-project';
 
 export interface RunRstestYamlProjectOptions {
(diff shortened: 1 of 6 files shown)


quanru:
已合并进 #2582:shared rstest runner 抽取(refactor(shared))已 cherry-pick 到 #2582 分支,并修复 @rstest/core/api 类型经 unknown 中转的 dts 构建问题。本 PR 不再单独维护。

PR #2537, opened 2026-05-25
Run YAML cases through Rstest

quanru:
## Summary
- Add a Rstest-backed framework runner for Midscene YAML cases.
- Route the existing `midscene <path>` YAML CLI flow through generated Rstest test files.
- Isolate real YAML execution in a child process so Rstest workers orchestrate pass/fail behavior without owning browser/runtime teardown.
- Preserve Midscene report and output summary generation, with focused framework unit coverage.

## Why
The UI testing framework needs YAML cases to participate in Rstest's runner model without replacing the repo's Vitest-based unit tests. Running Midscene/Puppeteer logic directly inside Rstest workers caused worker lifecycle issues, so the generated tests delegate the actual YAML run to a child process.

## Validation
- `pnpm run lint`
- `npx nx test @midscene/cli`
- `npx nx build @midscene/cli`

PR #2600, opened 2026-06-05
fix(cli): resolve @rstest/core from the CLI package, not process.argv[1]

quanru:
## Problem

Running YAML scripts fails with:

```
Cannot find module '@rstest/core/package.json'
```

even though `@rstest/core` is a declared dependency of `@midscene/cli`. The same logic written in TypeScript keeps working. Users hit this after rebuilding a Docker image, with no source change of their own.

## Root cause

Since 1.8.8 (`Run YAML cases through Rstest`, #2537), the YAML path resolves `@rstest/core` and `@rsbuild/core` at runtime via `require.resolve`. The lookup was anchored on **`process.argv[1]`** — the command-line entry — in `packages/cli/src/framework/rstest-runner.ts`:

```ts
const requireFromCliEntry = () => {
  const entry = process.argv[1] ? resolve(process.argv[1]) : join(process.cwd(), 'midscene-cli.js');
  return createRequire(entry);
};
```

When the launcher differs from the install location — a Docker entrypoint, a wrapper script, a relocated/symlinked bin, or an `npx` cache path — the `node_modules` chain walked up from `process.argv[1]` does not contain `@rstest/core`, so resolution throws. TypeScript scripts never touch this code path, which is why TS kept working while YAML broke.

## Fix

Anchor the lookup on **this module's own location** (`__dirname` of the bundled output) instead of `process.argv[1]`. `@rstest/core` and `@rsbuild/core` are direct dependencies of `@midscene/cli`, so they always sit on the CLI package's own resolution path, regardless of how the CLI is launched. ESM consumers of the bundled output (which have no `__dirname`) fall back to the previous behaviour.

## Reproduction & verification

Against the published **1.9.0**, launching the CLI through a wrapper located outside the install tree reproduces the exact error:

```
Error: Cannot find module '@rstest/core/package.json'
    at resolvePackageFromRstestCore (.../@midscene/cli/dist/lib/index.js)
```

With the `__dirname` anchor, the same invocation resolves and runs to the execution summary.

## Validation

- `npx biome check` on the changed files — clean
- `npx vitest --run tests/unit-test/framework/rstest-runner.test.ts -t anchor` — passing (new regression test: resolution succeeds even when `process.argv[1]` points outside the install tree)

Note: a full `nx build cli` currently fails in this workspace on an unrelated pre-existing `@midscene/harmony` declaration-resolution error (reproduces on a clean `main` checkout), not introduced by this change.

### DIFF
--- a/packages/cli/src/framework/rstest-runner.ts
+++ b/packages/cli/src/framework/rstest-runner.ts
@@ -10,7 +10,20 @@ export interface RunRstestYamlProjectOptions {
   stdio?: 'inherit' | 'pipe';
 }
 
-const requireFromCliEntry = () => {
+// `@rstest/core` and `@rsbuild/core` are direct dependencies of `@midscene/cli`,
+// so they always sit on the resolution path of the CLI's own files. Anchor the
+// lookup to this module's location (`__dirname` of the bundled output) rather
+// than `process.argv[1]`: the command-line entry can be a wrapper script, a
+// symlinked bin, an `npx` cache path, or some other launcher whose `node_modules`
+// chain does not include `@rstest/core`. Anchoring on `process.argv[1]` is what
+// caused YAML runs to fail with "Cannot find module '@rstest/core/package.json'"
+// in environments where the launcher differs from the install location.
+const requireFromCliPackage = () => {
+  if (typeof __dirname !== 'undefined') {
+    return createRequire(join(__dirname, 'index.js'));
+  }
+  // ESM consumers of the bundled output have no `__dirname`; fall back to the
+  // command-line entry so programmatic usage keeps working.
   const entry = process.argv[1]
     ? resolve(process.argv[1])
     : join(process.cwd(), 'midscene-cli.js');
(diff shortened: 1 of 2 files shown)


quanru:
## Reproduced & verified against a published prepatch

Triggered a prepatch from this branch → published **`@midscene/cli@1.9.1-beta-20260605030300.0`** (the `beta` dist-tag was later moved by an unrelated release, so pin the exact version, not `@beta`).

Same machine, same launcher-outside-the-install-tree scenario (mimics a Docker entrypoint / wrapper / relocated bin):

**A) Published latest `1.9.0` (unfixed)**
```
Welcome to @midscene/cli v1.9.0
Error: Cannot find module '@rstest/core/package.json'
    at resolvePackageFromRstestCore (.../@midscene/cli/dist/lib/index.js:3265:52)
```

**B) Prepatch `1.9.1-beta-20260605030300.0` (this fix)**
```
Welcome to @midscene/cli v1.9.1-beta-20260605030300.0
📊 Execution Summary:
   Total files: 1
```

The published prepatch bundle confirms the new anchor:
```js
const requireFromCliPackage = () => createRequire(join(__dirname, 'index.js'));
```
(the `typeof __dirname` guard is constant-folded away by the CJS bundler, which is expected — `__dirname` is always defined in the CJS output that the `midscene` bin runs).

EAGzzyCSL (approved):

PR #2644, opened 2026-06-10
Handle invalid Studio model env config

quanru:
## Summary

- Convert Studio model connection validation failures from thrown exceptions into structured config errors.
- Show invalid `MIDSCENE_MODEL_FAMILY` values inside the Model Env Config modal instead of letting the renderer crash.
- Allow long validation errors to wrap in the modal and keep the connectivity test disabled while the config is invalid.

## Root Cause

Studio reused `ModelConfigManager.getModelConfig()` during render-time validation. When a saved or typed env value such as `MIDSCENE_MODEL_FAMILY=1` was invalid, `getModelConfig()` threw synchronously, which escaped React rendering and blanked the Studio UI.

## Validation

- `pnpm --filter studio test -- tests/connectivity-env.test.ts tests/model-env-config-modal.test.tsx tests/model-env-storage.test.ts tests/studio-recorder-model-config.test.ts`
- `pnpm run lint`
- `pnpm exec nx build studio`

Notes: lint completed with an existing Biome warning under `.claude/worktrees/...`; the Studio build completed with existing sharp/photon bundle warnings.

### DIFF
--- a/apps/studio/src/renderer/components/ShellLayout/ModelEnvConfigModal.tsx
+++ b/apps/studio/src/renderer/components/ShellLayout/ModelEnvConfigModal.tsx
@@ -225,11 +225,10 @@ export function ModelEnvConfigModal({
     () => resolveModelConnection(envValues),
     [envValues],
   );
-  const configError =
-    'error' in resolvedConnection ? resolvedConnection.error : null;
   const validationError =
-    configError && !configError.startsWith('Missing required keys:')
-      ? configError
+    'error' in resolvedConnection &&
+    resolvedConnection.kind === 'invalid-config'
+      ? resolvedConnection.error
       : null;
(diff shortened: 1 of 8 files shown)


quanru:
Addressed in `ccca2dbfe`.

The model connection helper now returns structured error kinds, so UI validation can keep displaying invalid config errors without throwing. The recorder path only falls back for missing required keys; if a saved Studio env block has all required keys but an invalid value such as `MIDSCENE_MODEL_FAMILY=1`, it now propagates the validation error instead of silently using the visualizer/global config.

Added recorder coverage for both cases:
- incomplete saved env falls back to the visualizer config
- invalid saved env throws and does not fall back

EAGzzyCSL (approved):

PR #2646, opened 2026-06-10
feat(bdd): @midscene/bdd — AI-native BDD runner on cucumber-js

ScriptedAlchemy:
## What is this?

A new private package, `@midscene/bdd`, where you write UI tests as **standard Gherkin** and the AI executes every step — **no step-definition code required**. Real cucumber-js drives the run (CLI, reports, hooks, parallel workers); Midscene does the work.

```gherkin
Scenario: Failed login is reported everywhere
  Given the demo shop is open on the login page
  When I try to sign in as the "admin" user with a wrong password
  Then an error toast shows on the screen
  # @agent
  Then the server log contains a failed-login warning, per $check-logs
  # @no-ai
  Then the login attempt counter increments
```

## The three routing rules

| Rule | Trigger | What runs |
|---|---|---|
| **1. Midscene by default** | every plain statement | UI agent: `aiAct` for Given/When, fail-closed `aiAssert` for Then (`# @soft` warns instead of failing) |
| **2. Coding agent bailout** | `# @agent` comment or `$skill-name` token | general-purpose agent via core `callAI` (`codex login`, no API key — or any OpenAI-compatible endpoint), with `features/skills/*.md` loaded into the prompt; verdicts are fail-closed |
| **3. Classic BDD** | `# @no-ai` comment or `@no-ai` tag | your registered callback (`defineStep`/Given/When/Then from `@midscene/bdd`); unimplemented steps fail with a ready-to-paste snippet |

## Reusable flows — step definitions authored in Gherkin

A `@flow @param:role @returns:greeting` scenario named with a cucumber expression (`Scenario: I am logged in as {string}`) becomes declaratively callable from any feature file (`Given I am logged in as "admin"`): fresh variable scope, only declared returns flow back, depth cap 2, cross-file discovery, cucumber-style ambiguity errors. `I remember the price as "price"` captures values; later steps reference `<price>` (mechanical substitution — the model never sees placeholders).

## How it works

cucumber-js loads `@midscene/bdd/register` (one catch-all step definition — cucumber matching is keyword-agnostic) via a one-line preset:

```js
// cucumber.js
module.exports = require('@midscene/bdd/profile').defineProfile();
```

Step-level routing uses comment annotations because standard Gherkin has no step-level tags; everything else (Background, Outline, Rule, data tables, doc strings, tags, hooks, formatters, parallel mode) is untouched cucumber behavior. See the README's Conventions section for the standard-vs-extension breakdown.

## Try it

```bash
pnpm --filter @midscene/bdd build
cd packages/bdd/example
codex login   # or export MIDSCENE_MODEL_* env
npx cucumber-js
```

## Validation

- `npx nx test @midscene/bdd` — 172 tests (unit + a pipeline integration suite over real `.feature` fixtures with stub agents; no model/browser needed)
- `npx nx build @midscene/bdd` — dual CJS/ESM + dts clean; `pnpm run lint` clean
- `npx cucumber-js --dry-run` over the example: register/preset wiring verified through real cucumber (catch-all matches all steps, `not @flow` excludes flow scenarios)
- A 3-reviewer simplify pass (quality/perf/reuse) was applied before this PR: compile-once matchers, memoized AST indexes, shared context builder, isolated model config for the general agent

## Lineage

Concept validated in the closed POC #2639 (custom flow-IR); this rebuild follows the design session's conclusions: real cucumber-js instead of a bespoke parser, a new standalone package importing the Midscene SDK, and the three routing rules above. `packages/testing-framework` is untouched.

## Known follow-ups

- Live end-to-end run of the example with codex (works in the POC's architecture; not CI-gated here)
- Optional: cucumber-expression-based snippet generation, site docs (en/zh) once the API settles


packages/bdd/example/midscene.config.ts
```diff
@@ -0,0 +1,10 @@
+import { join } from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { defineBddConfig } from '@midscene/bdd';
+
+export default defineBddConfig({
+  uiAgent: {
+    type: 'web',
```
ottomao, line 7: We should support Android , iOS and any other agents here

ottomao: How about we set the config be like

```typescript
  target: {
    type: 'android',
    options: {
      deviceId: process.env.ANDROID_DEVICE_ID,
      androidAdbPath: process.env.ANDROID_ADB_PATH,
      autoDismissKeyboard: false,
    },
  },
  uiAgentOptions: {
    someConfigForAgent: 'bar',
    generateReport: true,
  },```

File: packages/bdd/example/midscene.config.ts [hunk 2]
```diff
@@ -0,0 +1,10 @@
+import { join } from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { defineBddConfig } from '@midscene/bdd';
+
+export default defineBddConfig({
```
ottomao, line 5: Let's built a general agent in like open-code

ScriptedAlchemy: Implemented in https://github.com/web-infra-dev/midscene/pull/2677: the general agent is now a real CLI coding agent — opencode by default, Codex via `generalAgent.type: 'codex' — spawned per `# @agent`/`$skill` step, replacing the callAI chat-completion path entirely. Both adapters reuse the Midscene model endpoint/key (MIDSCENE_MODEL_*) by default with zero extra config, overridable via `generalAgent.model`/`env`, and codex degrades gracefully to `codex login` when no endpoint is set.

packages/bdd/example/midscene.config.ts
ScriptedAlchemy: Done in 62767961 — `uiAgent` now takes a flat per-type union: `{ type: 'web' | 'android' | 'ios' | 'harmony' | 'computer' | 'interface', ... }`. Field names reuse the yaml env vocabulary (`deviceId`, `launch`, `displayId`, `module`/`export`/`param`) so configs translate 1:1, and a new `uiAgentOptions` mirrors the yaml `agent:` block (generateReport, reportFileName, groupName, cache, ...). Kept the `uiAgent` key rather than `target:` — it is the package-wide convention here, and core marks the yaml `target:` key as deprecated. Platform packages are lazy-imported and declared as optional peer deps, so the package still has no hard device dependencies; a missing one fails naming the package to install. Also added a `scope: 'scenario' | 'worker'` lifecycle knob so device targets can reuse one connection per worker instead of reconnecting every scenario.

yuyutaotao:
# Proposal to update schema

Main changes in the new proposal:

- We keep Gherkin as the readable test structure, but drop the classic Cucumber
  step-definition model.
- Reason: step definitions hide behavior behind pattern matching. In an
  AI-native framework, plain steps already mean "run with the UI Agent", so
  deterministic code should be explicit instead of matched implicitly.

Example:

```gherkin
# [fixture.user.createSignedInBuyer]
Given buyer fixture:
  """json
  { "id": "buyer-smoke-001", "role": "buyer" }
  """
```

This replaces `# [no-ai]` + Cucumber step definitions.

Reusable flows also move away from `@flow` / `@param:*`:

```gherkin
# [define: checkout.addProductToCart]
Scenario: Add product to cart
  When I search for "${productName}"
  And I add the first matching product to the cart

# [use: checkout.addProductToCart]
Given product:
  """json
  { "productName": "Trail Backpack" }
  """
```

Other notes:

- `# [agent]` stays for coding-agent steps.
- `# [soft]` stays for soft `Then` assertions.
- `# [guide: path]` is added for injecting UI/domain context.
- `$skill` references should live inside explicit `[agent]` steps, instead of
  implicitly changing routing.

Overall: less Cucumber magic, more explicit Midscene routing.

yuyutaotao:
Here is some commen about principles

# BDD v2 Design Principles

Midscene BDD v2 uses Gherkin for readable UI scenarios, but does not inherit
classic Cucumber step definitions. A `.feature` file should show both the
business intent and the execution route of non-default steps.

```text
.feature = business scenario + lightweight Midscene annotations
.ts/.js  = deterministic fixture and project logic
.md      = AI context
AI       = natural-language UI executor
```

## Core Ideas

### Keep Gherkin As The Scenario Language

`Feature`, `Scenario`, `Given`, `When`, and `Then` remain the main authoring
surface. They give product, QA, and engineering a shared format for reviewing
behavior and locating failures.

Plain steps are UI Agent steps by default:

```gherkin
When I open the cart page
Then the cart line item shows quantity controls
```

### Make Non-Default Execution Explicit

Classic Cucumber step definitions hide behavior behind pattern matching. In
Midscene, plain natural-language steps already have a default executor, so
deterministic code and coding-agent work should be visible in the feature file.

Use comment annotations for that routing:

```gherkin
# [agent]
Then verify the latest order in $database

# [fixture.user.createSignedInBuyer]
Given buyer fixture:
  """json
  { "id": "buyer-smoke-001", "role": "buyer" }
  """
```

### Keep Annotations Small And Purposeful

The initial annotation set is intentionally small:

- `# [guide: path]` adds AI context from Markdown.
- `# [agent]` routes one step to the coding agent.
- `# [soft]` makes the next `Then` a soft assertion.
- `# [fixture.name]` calls deterministic project code.
- `# [define: name]` marks a reusable scenario.
- `# [use: name]` runs a reusable scenario.

Avoid adding loops, conditionals, feature-level variables, or broad import
systems to `.feature` files.

### Use Markdown Only As AI Context

Guides describe page knowledge, domain terms, and operation hints. They do not
execute and should not become hidden flows.

```gherkin
# [guide: guides/product-search.md]
When I search for "running shoes"
```

### Keep Deterministic Logic In TypeScript

`fixture.*` annotations are for explicit, deterministic work such as data
setup, cleanup, database checks, or project-specific scripts. The fixed
`fixture.` prefix keeps top-level annotation names available for framework
features.

Complex parsing, branching, loops, and API integration belong in TypeScript,
not in Gherkin.

### Allow Reusable Scenarios, With Limits

`define/use` is the escape hatch for report-visible UI reuse when repeating
steps would hurt readability.

```gherkin
# [define: checkout.addProductToCart]
Scenario: Add product to cart
  When I search for "${productName}"
  And I add the first matching product to the cart

# [use: checkout.addProductToCart]
Given product:
  """json
  { "productName": "Trail Backpack" }
  """
```

Reusable scenarios should stay shallow, avoid recursion, and remain UI-focused.
Stable project logic should still move to `fixture.*` TypeScript.

## Project Shape

```text
midscene.config.ts
features/
  checkout.feature
fixtures/
  user/createSignedInBuyer.ts
  order/captureLatestOrder.ts
guides/
  checkout-ui.md
  order-domain.md
```

## Routing Order

1. `# [fixture.name]` runs deterministic TypeScript.
2. `# [agent]` runs the configured coding agent.
3. `# [use: name]` expands a reusable scenario.
4. Plain steps run with the UI Agent.

`# [soft]` modifies assertion behavior. `# [guide: path]` contributes context.

## Non-Goals

- No classic Cucumber step-definition API.
- No `@flow` or `@param:*`.
- No tag-driven execution semantics.
- No loops, conditionals, or feature-level variables in `.feature`.
- No Markdown-as-flow behavior.

ScriptedAlchemy:
Closing in favor of #2706. This draft explored the broader cucumber-js/@midscene/bdd design and remains useful as historical reference, but #2706 intentionally ships the smaller CLI .feature runner first and complements the merged core runGherkinScenario API from #2697.

Issue #1052, opened 2025-08-13
[Feature]: config 配置需要使用登录concurrent设置>1 问题


aceHubert:
### What problem does this feature solve?

<img width="819" height="384" alt="Image" src="https://github.com/user-attachments/assets/65f563c4-bd7f-4215-8ca9-51ecc886bd78" />
concurrent 怎么在 login 之后有效啊？

### What does the proposed API look like?

把files 拆分成post_xxx 可单独配置

PR #2710, opened 2026-06-24
feat(cli): run a setup script before parallel yaml tasks

### LINKED ISSUE #2031: [Feature]: Support serial pre-tasks (e.g. login) + parallel yaml tasks with shared browser context
### Feature Description

In real-world automation scenarios, we often have **stateful pre-tasks** (e.g. login)
followed by **multiple independent business scenarios** that can safely run in parallel
*after* the authenticated state is established.

Midscene YAML runner currently supports:

- `concurrent > 1` for parallel execution
- `shareBrowserContext: true` for sharing cookies / storage

However, these two options conflict when YAML scripts have **implicit dependencies**.

---

### Problem

Consider the following YAML scripts:

- `login.yaml` — authenticate user and establish session
- `search.yaml` — search data after login
- `report.yaml` — view or generate reports after login
- `settings.yaml` — read or update user settings after login

When running them together with:

```yaml
shareBrowserContext: true
concurrent: 3
```

the actual behavior is:

All YAML scripts start at the same time

search.yaml, report.yaml, or settings.yaml may execute
before login is completed

This leads to nondeterministic failures, such as:

- being redirected to login page
- missing permissions
- inconsistent assertions

### What does the proposed API look like?

```yaml
setupFiles:
  - login.yaml

files:
  - search.yaml
  - report.yaml
  - settings.yaml

shareBrowserContext: true
concurrent: 3
```

quanru:
## Background

Resolves #2031.

The YAML batch runner already supports `concurrent > 1` and `shareBrowserContext: true`, but they conflict when scripts have an implicit dependency: every script starts at once, so a login script may finish *after* the business scripts that depend on it.

This PR adds a `setup` script that runs **before** the parallel main scripts, while sharing the same browser context.

## What changed

- **`setup` config option** (`MidsceneYamlConfig` + `BatchRunnerConfig`): a single YAML script executed before the main `files`. A glob/relative path is supported but must resolve to exactly one file.
- **Shared context**: the setup and main scripts reuse the same `browser` / shared `page`, so state established during setup (e.g. a login) is visible to every main script.
- **Fail-fast semantics**: a setup failure aborts the whole batch; the failing setup is recorded as `failed` and all main scripts are reported as `notExecuted`. This holds regardless of `continueOnError`, since the main scripts rely on the prerequisite state.
- **Validation, enforced at two layers**:
  - Config layer (`config-factory`): `setup` requires `shareBrowserContext: true`, otherwise it throws — the setup result cannot reach the main scripts without a shared context.
  - Executor layer (`YamlBatchExecutor`): the same invariant is re-checked, plus a guard that rejects a YAML file used as both the setup and a main file (which would otherwise reuse one already-finished player keyed by resolved path).
- **CLI flag `--setup <file>`**: a single setup script on the command line.
- **Execution plan output**: the setup script is listed separately as `Setup script:`.

### Example — config file

```yaml
setup: ./scripts/login.yaml

files:
  - ./scripts/search.yaml
  - ./scripts/report.yaml
  - ./scripts/settings.yaml

shareBrowserContext: true
concurrent: 3
```

### Example — command line

```bash
midscene \
  --setup ./scripts/login.yaml \
  --files ./scripts/search.yaml ./scripts/report.yaml ./scripts/settings.yaml \
  --share-browser-context \
  --concurrent 3
```

## Note on rstest

This "shared context + parallel" path is driven by the CLI's own `p-limit` batch executor (rstest only hosts the single batch test), so the setup-then-parallel orchestration lives entirely in our code — rstest was not touched.

## Tests

- `config-factory.test.ts`: `setup` resolution order, undefined when absent, error when it matches zero or multiple files, and the `shareBrowserContext` validation (pass + throw) for both `createConfig` and `createFilesConfig`.
- `batch-runner.test.ts`: setup runs before the main scripts and all succeed; a setup failure aborts and leaves the main scripts not executed; the executor throws when `setup` is set without `shareBrowserContext`; the executor throws when a file is both the setup and a main file.
- `cli-utils.test.ts`: `--setup` argument parsing.

## Docs

- `apps/site/docs/{en,zh}/yaml-script-runner.mdx`: documented the `--setup` flag and the setup-then-parallel config example (bilingual).

## Validation

- `pnpm run lint` — clean (1263 files)
- `npx nx test cli` — 191 passed
- `npx nx build cli` — passed (type-checked)


yuyutaotao (approved):

quanru:
close https://github.com/web-infra-dev/midscene/issues/1052

PR #2720, opened 2026-06-25
feat(ios): source frameSequence frames from WDA MJPEG stream

quanru:
## Background

`frameSequence` works well on Android because its `screenshotBase64()` pulls the latest frame from a continuously-running **scrcpy** video stream (near-instant). On **iOS**, `screenshotBase64()` goes through WDA's `takeScreenshot()` — a full per-call capture that is too slow to sample a short time window densely, so the effective frame spacing becomes `captureLatency + intervalMs` and short-lived UI (toasts) slips through the gaps.

Key point: iOS **already runs WDA's native MJPEG server** (`mjpegStreamUrl`, with `mjpegServerFramerate`/`mjpegServerScreenshotQuality` configured) — the iOS analog of scrcpy. It just wasn't wired into frame capture. This PR consumes it as a fast frame source.

Targets `feat/frame-sequence-transient-ui` (the frameSequence feature branch).

## What this PR does

- **core (device)**: add an optional `AbstractInterface.captureFrameSequence({count, intervalMs, abortSignal})` — a fast multi-frame capture implemented by devices that maintain a continuous frame stream. Returns data-URL frames in temporal order; may return fewer than requested.
- **core (agent)**: `captureUIContextSequence` prefers `captureFrameSequence` for the earlier frames and captures the representative (last) frame with a normal full-quality screenshot. If the fast source throws or yields nothing, it falls back to the existing sequential `screenshotBase64()` loop. Abort is honored throughout.
- **ios**: `MjpegFrameSource` consumes the WDA MJPEG stream (lazy-started, auto-reconnecting with backoff, stopped on `destroy()`) and keeps the latest decoded JPEG. `IOSDevice.captureFrameSequence` samples it at the requested cadence — pulling is near-instant, so frames land at `intervalMs` instead of `intervalMs + slowScreenshot`.

`screenshotBase64()` (normal locate/asserts) is unchanged — it still uses full-quality `takeScreenshot()`. The MJPEG stream is only used for the multi-frame sampling.

## Why this design

- Mirrors what already makes Android fast (continuous stream → instant latest-frame), instead of optimizing the slow per-call path.
- The fast path is an **optional** capability: Android/web don't implement it and keep working (Android is already fast; web could later implement it over CDP screencast). No behavior change for devices without it.
- Graceful degradation: any stream failure falls back to sequential capture, so correctness is preserved even if the MJPEG server is unavailable.

## Tests

- `packages/ios/tests/unit-test/mjpeg-frame-source.test.ts` — `extractJpegFrames` (single/multiple frames, boundary/header bytes, incomplete frame, split-across-chunks, trailing-FF), plus a streamed `MjpegFrameSource` decoding the latest frame from a mocked fetch stream. 7/7 pass.
- `packages/core/tests/unit-test/agent-frame-sequence.test.ts` — fast-path is used when the device exposes `captureFrameSequence` (earlier frames from stream + one representative screenshot), and falls back to sequential capture when it fails.

## Validation

- `npx tsc --noEmit` for `@midscene/core` and `@midscene/ios` — clean
- `npx nx build core` and `npx nx build ios` — pass
- Core frameSequence suites (20 tests) + iOS `device.test.ts` (54) + new MJPEG tests — pass
- biome lint clean

Note: end-to-end behavior was verified on Android/web in the base PR; the iOS path here is covered by unit tests (no physical iOS device in CI).

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -459,6 +459,46 @@ export class Agent<
     const { count, intervalMs } =
       this.normalizeFrameSequenceOption(frameSequence);
 
+    // Fast path: devices with a continuous frame stream (e.g. iOS WDA MJPEG)
+    // sample recent frames far faster than repeated screenshotBase64() calls.
+    // We pull the earlier frames from the stream and capture the representative
+    // (last) frame with a normal full-quality screenshot.
+    const fastCapture = this.interface.captureFrameSequence?.bind(
+      this.interface,
+    );
+    if (fastCapture) {
+      try {
+        const earlier = await fastCapture({
+          count: count - 1,
+          intervalMs,
+          abortSignal,
+        });
+        abortSignal?.throwIfAborted();
+        const representative = await this.getUIContext('assert');
+        const frames = [
+          ...earlier.map((frame) =>
+            ScreenshotItem.create(frame.base64, frame.capturedAt),
+          ),
+          representative.screenshot,
+        ];
+        if (frames.length > 1) {
+          return {
+            ...representative,
+            screenshotSequence: frames,
+          };
+        }
+        // Stream yielded no earlier frames; fall through to sequential capture.
+      } catch (error) {
+        if (abortSignal?.aborted) {
+          throw error;
+        }
+        debug(
+          'frame sequence fast path failed, falling back to sequential capture: %s',
+          error,
+        );
+      }
+    }
+
     const frames: ScreenshotItem[] = [];
     let lastContext: UIContext | undefined;
     for (let i = 0; i < count; i++) {
--- a/packages/core/src/device/device-options.ts
+++ b/packages/core/src/device/device-options.ts
@@ -141,6 +141,17 @@ export type IOSDeviceOpt = {
   useWDA?: boolean;
   /** WDA MJPEG server port for real-time screen streaming (default: 9100) */
   wdaMjpegPort?: number;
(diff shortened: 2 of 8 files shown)


quanru:
Follow-up (7d8101dff): the WDA MJPEG frame source is now **opt-in, disabled by default**, mirroring Android scrcpy.

- Enable via `new IOSDevice({ wdaMjpegFrameSource: { enabled: true }, wdaMjpegPort })`. When off (default), `frameSequence` behaves exactly as before — sequential `screenshotBase64()` capture — so there is no behavior change unless explicitly enabled.
- The capability (`captureFrameSequence`) is only attached when enabled, so the agent transparently falls back when it is off.
- **Multi-device concurrency**: the stream URL is built from the existing per-device `wdaMjpegPort` option, so each device streams from its own port (pass a distinct `wdaMjpegPort` per device, just like `wdaPort`). WDA-side MJPEG port forwarding is the user's responsibility, same as the WDA control port.

Added tests for the opt-in gating and the per-device stream URL.

quanru:
Superseded by #2753, which fully incorporates this PR's WDA MJPEG frame source implementation (commits 3bf66f12 and 7d8101df are both included) and extends it: the iOS MJPEG stream is now exposed via the unified `openFrameSource()` interface (replacing the one-shot `captureFrameSequence`), so it powers the new `UIObserver` alongside Android scrcpy and Web CDP.

PR #2726, opened 2026-06-27
fix(playground): respect screenshotIncluded request param for dump/report

### LINKED ISSUE #2539: [Bug]: Playground /execute endpoint returns oversized response causing RangeError: Invalid string length
### Version

```sh
"@midscene/playground": "^1.8.4"
"@midscene/android": "^1.8.4"
"@midscene/shared": "^1.8.4"
```

### Details

Details
When calling the /execute endpoint, the playground server constructs a response with dump and reportHTML fields that include base64-inlined screenshots (inlineScreenshots: true). After a case with multiple actions executes, the response object becomes extremely large (>50MB), causing JSON.stringify to throw RangeError: Invalid string length inside ServerResponse.json().

Stack trace:

RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at ServerResponse.json
    at .../dist/lib/server.js:774-783

Root cause found in source code:

node_modules/@midscene/playground/dist/lib/server.js lines 774-783:

const dumpString = agent.dumpDataString({
    inlineScreenshots: true
});
// ...
response.reportHTML = agent.reportHTMLString({
    inlineScreenshots: true
});

inlineScreenshots: true is hardcoded, not configurable via request parameters. Even when sending screenshotIncluded: false in the request body, the dump and reportHTML fields still embed base64 screenshots, making the response grow linearly with the number of executed actions.

### Reproduce link

Related issues #1439 (same error, closed without fix)

### Reproduce Steps

Any /execute call with a case that performs many actions (e.g., multiple swipe operations) will eventually trigger this error as the dump field accumulates screenshot data.

zhsj0089944:
## Problem

The `/execute` endpoint hardcodes `inlineScreenshots: true` when generating `dump` and `reportHTML` response fields. After multiple actions, the response can exceed 50MB, causing `JSON.stringify` to throw `RangeError: Invalid string length`.

The `screenshotIncluded` field is already destructured from the request body and respected during action execution, but ignored when building the response.

## Fix

Respect the `screenshotIncluded` parameter from the request body when generating dump and reportHTML:

- When `screenshotIncluded: false` is sent, dump and reportHTML are generated **without** inline base64 screenshots, keeping the response size manageable.
- When `screenshotIncluded` is not provided (or `true`), the existing behavior is preserved (default `true` for backward compatibility).

## Related Issue

Fixes #2539

### DIFF
--- a/packages/playground/src/server.ts
+++ b/packages/playground/src/server.ts
@@ -2790,8 +2790,9 @@ class PlaygroundServer {
       }
 
       try {
+        const inlineScreenshots = screenshotIncluded !== false;
         const dumpString = agent.dumpDataString({
-          inlineScreenshots: true,
+          inlineScreenshots,
         });
         if (dumpString) {
           const groupedDump = ReportActionDump.fromSerializedString(dumpString);
@@ -2800,7 +2801,7 @@ class PlaygroundServer {
           response.dump = null;
         }
         response.reportHTML =
-          agent.reportHTMLString({ inlineScreenshots: true }) || null;
+          agent.reportHTMLString({ inlineScreenshots }) || null;
 
         agent.writeOutActionDumps();
         agent.resetDump();


quanru:
Thanks for the contribution and for identifying the oversized response problem.

We are closing this PR because the proposed serialization change would make `dump` and `reportHTML` contain screenshot references without returning the corresponding image assets. The Playground browser consumes these response fields directly, so replay and report previews can lose their screenshots when `screenshotIncluded` is `false`.

The `/execute` response path on `main` has since been redesigned to register and return the incrementally persisted report instead of synchronously embedding `dump` and `reportHTML`. That newer flow addresses the large-response and main-thread blocking concerns without leaving unresolved screenshot references, so this patch is now obsolete.

Thanks again for reporting the original issue and proposing a focused fix.

PR #2746, opened 2026-07-01
fix(core): gpt.ts - for gpt re-scale coordinates from model coordinat…

### LINKED ISSUE #2745: [Bug] Issues with screen resolution in computer-use-mode with gpt-5
Hi,
I am new to midscene.js and it might not be the right place for general usage questions. If this is not right place (as there are only bugs and features in this list), let me know where to place the question.

I am using midscene.js in computer-use mode on Linux (Debian trixie). When using this yaml file:

computer: {
  screenshotShrinkFactor:1
}

tasks:
  - name: Open Nautilus
    flow:
      - aiAct: Press the meta key
      - aiAct: Click on Nautilus icon to launch the file manager.

It never hits the Nautilus-Icon in the dash. The debug log is showing:

{
  "locate": {
    "prompt": "Nautilus file manager icon in the bottom dock",
    "bbox": [324, 704, 358, 742]
  }
}

Which is obvious not correct. THe dash/dock is at the bottom of the screen which as a resolution of 1920x1080.

What do I need to do to force midscene or better the llm to use the correct screen resolution?

Thanks for your help
Michael

miti4711:
…e space to physical screen size.

- handles issues when using gpt5 and computer-use and screen resolutions other than 1366x768.
- for gpt coordinates are re-scaled between model coordinates system and physical screen coordinates system

Info for Reviewer: the issue has been fixed with the help of gpt5-5. This is my first contribution. So let me know what to do better.

fixes #2745

### DIFF
--- a/packages/core/src/ai-model/models/gpt.ts
+++ b/packages/core/src/ai-model/models/gpt.ts
@@ -1,10 +1,16 @@
 import type { TModelFamily } from '@midscene/shared/env';
+import { zoomForGPT4o } from '@midscene/shared/img';
 import type {
   ChatCompletionCallContext,
   ChatCompletionParamsResult,
   ImageDetail,
   ModelAdapterDefinition,
 } from '../model-adapter/types';
+import type {
+  LocateResultContext,
+  LocateResultValue,
+  PixelBbox,
+} from '../shared/model-locate-result';
 
 const originalImageDetailForDefaultIntent = (
   input: ChatCompletionCallContext,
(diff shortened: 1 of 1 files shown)


EAGzzyCSL:
Thanks for the contribution and for looking into this issue.

I don’t think this is the right direction to fix it, though. The main concern is that this implementation tries to infer GPT’s internal image scaling behavior with `zoomForGPT4o()`. That is a heuristic, not a stable contract from the GPT-5 API, and it may become inaccurate across model versions or OpenAI-compatible providers. It can also accidentally rescale coordinates that were already returned in the original pixel space.

Since we already know GPT works better with certain screenshot sizes, I think the more reliable approach is to control the input image size before sending it to the model. Midscene already supports this through `screenshotShrinkFactor`, which shrinks the screenshot and keeps the coordinate conversion path explicit and testable.

So instead of adding GPT-specific coordinate rescaling in the result adapter, I’d suggest configuring `screenshotShrinkFactor` for GPT-5 usage to keep screenshots closer to the recommended dimensions. This also matches the guidance in our GPT-5 model configuration docs.

PR #2837, opened 2026-07-21
fix(core): compact duplicate report dumps on finalize

quanru:
## Summary

- Compact superseded grouped dump scripts when `ReportGenerator` finalizes a report.
- Preserve the latest execution for each stable execution ID while retaining all legacy executions without IDs.
- Keep live report generation append-only and preserve inline and directory-mode screenshot data.
- Rewrite through a sibling temporary file and replace the finalized report atomically.

## Root cause

Report generation appends a complete dump snapshot on every execution update, and `finalize()` appends the latest execution once more to capture its final state. The report UI deduplicates these snapshots when loading the page, but the HTML file retained every historical dump.

In the supplied report, this produced 487 grouped dump tags and inflated the file from the compacted 94.94 MiB to 150.32 MiB, an increase of approximately 58.33%.

## Impact

Completed reports now contain one compacted dump per report group without changing append-only behavior while a run is active.

The supplied report was reduced from 157,617,354 bytes to 99,547,590 bytes (36.84% smaller). Browser-level verification confirmed that the merged dump content and image index remained unchanged: 1 execution, 119 tasks, and 83 image nodes.

## Regression coverage

The regression test was added before the implementation and failed with three dump tags instead of one. It now passes and also covers:

- multiple execution IDs
- legacy executions without IDs
- inline screenshot mode
- directory screenshot mode
- report merging performance

## Validation

- `pnpm run lint`
- `pnpm --dir packages/core exec vitest --run tests/unit-test/report-generator.test.ts tests/unit-test/report-merge-count.test.ts` — 57 tests passed
- `pnpm exec nx build @midscene/core`
- Real-report compaction — 487 dump tags reduced to 1 in 1.15 seconds with equivalent browser-visible content
- `pnpm exec nx test @midscene/core` — 1,340 passed, 8 skipped, and 1 unrelated size-threshold failure in `merge-browser-parse.test.ts` (expected less than 15 MiB; current generated report is 16.92 MiB)

### DIFF
--- a/packages/core/src/dump/report-dump-compactor.ts
+++ b/packages/core/src/dump/report-dump-compactor.ts
@@ -1,0 +1,321 @@
+import { randomUUID } from 'node:crypto';
+import { createReadStream } from 'node:fs';
+import { type FileHandle, open, rename, stat, unlink } from 'node:fs/promises';
+import { basename, dirname, join } from 'node:path';
+import { antiEscapeScriptTag, escapeScriptTag } from '@midscene/shared/utils';
+import { type ExecutionDump, ReportActionDump } from '../types';
+
+const SCRIPT_OPEN = Buffer.from('<script');
+const SCRIPT_CLOSE = Buffer.from('</script>');
+const READ_CHUNK_SIZE = 32 * 1024 * 1024;
+const DUMP_TYPE_ATTRIBUTE = 'type="midscene_web_dump"';
+const GROUP_ID_REGEXP = /data-group-id="([^"]+)"/;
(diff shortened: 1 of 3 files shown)


quanruzhuoxiu:
Closing this PR only to recreate it under the correct GitHub account, @quanruzhuoxiu. A replacement PR link will be added shortly.

quanruzhuoxiu:
Replacement PR: #2838. It contains the same branch and commit, and was created by @quanruzhuoxiu.

PR #2838, opened 2026-07-21
fix(core): compact duplicate report dumps on finalize

quanruzhuoxiu:
## Summary

- Compact superseded grouped dump scripts when `ReportGenerator` finalizes a report.
- Preserve the latest execution for each stable execution ID while retaining all legacy executions without IDs.
- Keep live report generation append-only and preserve inline and directory-mode screenshot data.
- Rewrite through a sibling temporary file and replace the finalized report atomically.

## Root cause

Report generation appends a complete dump snapshot on every execution update, and `finalize()` appends the latest execution once more to capture its final state. The report UI deduplicates these snapshots when loading the page, but the HTML file retained every historical dump.

In the supplied report, this produced 487 grouped dump tags and inflated the file from the compacted 94.94 MiB to 150.32 MiB, an increase of approximately 58.33%.

## Impact

Completed reports now contain one compacted dump per report group without changing append-only behavior while a run is active.

The supplied report was reduced from 157,617,354 bytes to 99,547,590 bytes (36.84% smaller). Browser-level verification confirmed that the merged dump content and image index remained unchanged: 1 execution, 119 tasks, and 83 image nodes.

## Regression coverage

The regression test was added before the implementation and failed with three dump tags instead of one. It now passes and also covers:

- multiple execution IDs
- legacy executions without IDs
- inline screenshot mode
- directory screenshot mode
- report merging performance

## Validation

- `pnpm run lint`
- `pnpm --dir packages/core exec vitest --run tests/unit-test/report-generator.test.ts tests/unit-test/report-merge-count.test.ts` — 58 tests passed
- `pnpm exec nx build @midscene/core`
- Real-report compaction — 487 dump tags reduced to 1 in 1.15 seconds with equivalent browser-visible content
- `pnpm exec nx test @midscene/core` — 1,340 passed, 8 skipped, and 1 unrelated size-threshold failure in `merge-browser-parse.test.ts` (expected less than 15 MiB; current generated report is 16.92 MiB)

## Previous review history

Supersedes #2837, which was recreated so the pull request and commits are both attributed to @quanruzhuoxiu. The previous automated inline review remains available on #2837.

## CI follow-up

- Replaced the Node-only `randomUUID` import with the browser-compatible shared `uuid()` helper so browser bundles can link the report compactor.
- `pnpm --dir packages/core exec vitest --run tests/unit-test/report-generator.test.ts tests/unit-test/report-merge-count.test.ts` — 58 tests passed
- `pnpm run lint`
- `pnpm exec nx build computer-playground --skip-nx-cache`
- `pnpm exec nx run-many --target=build --exclude=doc --verbose --skip-nx-cache` — 27 projects built successfully

## Review follow-up

- Treat final report compaction as best-effort so an `ENOSPC` or other compaction error cannot turn a completed Agent run into a failure.
- Preserve and return the uncompressed report while logging a Midscene warning.
- Added a regression test that forces compaction to reject and verifies `finalize()` still resolves with the report path.

### DIFF
--- a/packages/core/src/dump/report-dump-compactor.ts
+++ b/packages/core/src/dump/report-dump-compactor.ts
@@ -1,8 +1,11 @@
-import { randomUUID } from 'node:crypto';
 import { createReadStream } from 'node:fs';
 import { type FileHandle, open, rename, stat, unlink } from 'node:fs/promises';
 import { basename, dirname, join } from 'node:path';
-import { antiEscapeScriptTag, escapeScriptTag } from '@midscene/shared/utils';
+import {
+  antiEscapeScriptTag,
+  escapeScriptTag,
+  uuid,
+} from '@midscene/shared/utils';
 import { type ExecutionDump, ReportActionDump } from '../types';
 
 const SCRIPT_OPEN = Buffer.from('<script');
@@ -257,7 +260,7 @@ async function replaceReportFile(
 
   const backupPath = join(
     dirname(reportPath),
-    `.${basename(reportPath)}.${randomUUID()}.backup`,
+    `.${basename(reportPath)}.${uuid()}.backup`,
   );
   await rename(reportPath, backupPath);
   try {
@@ -293,7 +296,7 @@ export async function compactReportDumps(
 
   const temporaryPath = join(
     dirname(reportPath),
-    `.${basename(reportPath)}.${randomUUID()}.tmp`,
+    `.${basename(reportPath)}.${uuid()}.tmp`,
   );
 
   try {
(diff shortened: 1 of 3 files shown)


quanruzhuoxiu:
CI follow-up: commit `35cca4460` replaces the Node-only `randomUUID` import with the browser-compatible shared `uuid()` helper. The replacement AI unit test run [29828013609](https://github.com/web-infra-dev/midscene/actions/runs/29828013609) completed successfully, including both `Build project` and `Run tests with coverage`. A subsequent best-effort compaction follow-up commit (`dbef91433`) has triggered a fresh CI cycle on the latest head.

EAGzzyCSL (approved):

PR #2855, opened 2026-07-23
feat(core): make screenshot lifecycle format-aware

quanruzhuoxiu:
## Summary

- make `ScreenshotItem` retain its detected image format instead of assuming PNG
- persist screenshot references with format-aware file extensions and MIME types
- update report action dumps and screenshot storage to preserve WebP bytes end to end
- add regression coverage for screenshot creation, storage, serialization, shrink factors, and orientation handling

## Stack

This is PR 2 of 4 for the WebP foundation split.

- Depends on #2854
- The report/export consumer layer follows this PR
- The original foundation PR #2851 remains unchanged as a rollback option

## Validation

- `pnpm --filter @midscene/core test` (124 files, 1,349 passed, 8 skipped)
- `pnpm exec nx build @midscene/core`
- `pnpm run type-check:tests`
- `pnpm run lint`

PR #2854, opened 2026-07-23
feat(shared): add WebP image primitives

quanruzhuoxiu:
## Summary

- centralize image format detection, MIME/extension mapping, and base64 parsing for PNG, JPEG, and WebP
- add a browser-safe WebP encoder and integrate it with shared image transforms
- preserve source formats through crop, scale, and image-info operations
- add focused unit coverage for WebP detection, conversion, fallback behavior, and transforms

## Stack

This is PR 1 of 4 for the WebP foundation split and is based on `main`.

The original foundation PR #2851 remains unchanged as a rollback option.

## Validation

- `pnpm --filter @midscene/shared test` (32 files, 431 tests passed)
- `pnpm exec nx build @midscene/shared`
- `pnpm run lint`

PR #2851, opened 2026-07-23
feat(shared): establish canonical WebP screenshot pipeline

quanruzhuoxiu:
## Summary

- add a shared PNG/JPEG/WebP format contract and canonicalize static screenshots to WebP
- use Sharp in Node.js and Canvas/OffscreenCanvas in browsers while preserving native WebP and explicitly preserved JPEG streams
- propagate MIME type and file extension through screenshot identity, storage, HTML reports, split reports, Markdown export, Report App, and Visualizer
- retain the exact post-processing image bytes sent to the model and associate them with report evidence by SHA-256
- document WebP input support in both English and Chinese API references

## PR stack

This is PR 1 of 3 and targets `main`.

The browser-producer and device-producer PRs will target this branch. The original PR #2826 and `feat/chrome-extension-default-webp` branch remain unchanged as a rollback option.

## Validation

- `pnpm --filter @midscene/shared test` (33 files, 433 tests)
- `pnpm --filter @midscene/core test` (125 files, 1357 passed, 8 skipped)
- `pnpm --filter @midscene/report test` (16 files, 87 tests)
- `pnpm --filter @midscene/visualizer test` (21 files, 104 tests)
- `pnpm exec nx run-many -t build -p @midscene/shared @midscene/core @midscene/report @midscene/visualizer`
- `pnpm exec nx run-many -t build -p @midscene/android-playground @midscene/computer-playground @midscene/ios @midscene/harmony`
- `pnpm run type-check:tests`
- `pnpm run lint`

PR #2856, opened 2026-07-23
feat(core): record exact model input screenshots

quanruzhuoxiu:
## Summary

- record the exact image bytes passed to each model request after preprocessing
- annotate model-input screenshots with a SHA-256 digest for report verification
- reuse existing screenshot objects when the model receives identical bytes
- deduplicate retry recordings while preserving parent runtime callbacks

## Stack

This is PR 3 of 4 for the WebP foundation split. It is intentionally based on `main` and can be reviewed and merged independently from #2854 and #2855.

The original foundation PR #2851 remains unchanged as a rollback option.

## Validation

- `pnpm exec nx run-many -t build -p @midscene/shared @midscene/core`
- `pnpm --filter @midscene/core test` (124 files, 1,347 passed, 8 skipped)
- `pnpm run type-check:tests`
- `pnpm run lint`

### DIFF
--- a/packages/core/src/agent/model-input-recorder.ts
+++ b/packages/core/src/agent/model-input-recorder.ts
@@ -1,0 +1,62 @@
+import { Buffer } from 'node:buffer';
+import type { ModelRuntime } from '@/ai-model/models';
+import { ScreenshotItem } from '@/screenshot-item';
+import type { ExecutionTask } from '@/types';
+import { parseBase64 } from '@midscene/shared/img';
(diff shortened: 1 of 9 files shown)


quanru:
Closing this outdated implementation. The final model image bytes are now determined at the model-request preparation boundary, while this PR hooks older task-level call sites and no longer guarantees that the recorded image matches the payload sent to the provider. Exact request payloads can already be inspected through model-call recording. If report-visible model input evidence is still needed, it should be redesigned on top of the current request pipeline rather than rebasing this implementation.

PR #2870, opened 2026-07-24
feat(shared): support reference images in the act CLI tool

codeestX:
## What

Adds reference-image input to the `act` common tool, so platform CLIs
(`@midscene/web`, `@midscene/android`, `@midscene/ios`, …) can pass images to
`act` the same way they already can to `assert`:

```bash
npx @midscene/web act \
  --url https://example.com \
  --prompt "tap the icon that matches the reference image" \
  --image ./target.png \
  --image-name target
```

`--image` / `--image-name` may be repeated (counts must match) and remote URLs
can be inlined via `--convert-http-image2-base64 true`.

## Why

`act` was the odd one out: only `assert` exposed the image flags at the CLI
surface, even though core already supports multimodal prompts everywhere.
`aiAct` / `aiAction` accept a `TUserPrompt` and forward reference images to the
planner (`TaskExecutor.action` → `planImpl({ referenceImageMessages })`,
`packages/core/src/agent/tasks.ts`). So this is purely CLI wiring — no core
change was needed.

## How

- add `promptInputExtraSchema` (`image` / `imageName` / `convertHttpImage2Base64`)
  to the `act` tool schema
- compose a `UserPromptLike` via `composeUserPrompt(...)` before calling
  `aiAction`, mirroring `assert`
- widen `BaseAgent.aiAction` to accept `UserPromptLike`

Plan-cache correctness is preserved: the plan cache key is
`JSON.stringify(prompt)`, which already incorporates each image name/url, so
"same text, different image" does not hit a stale cached plan.

`aiWaitFor` intentionally stays string-only (no multimodal entry point yet); the
stale comment in `types.ts` is updated accordingly.

## Validation

- `npx nx test @midscene/shared -- tests/unit-test/tool-generator.test.ts` — 45 passed
  (new `generateCommonTools — act image prompts` cases: forwards images as a
  TUserPrompt object, forwards a local-path url verbatim, no-image prompt stays a
  bare string; updated schema assertions)
- `npx nx build @midscene/shared` — passed with the type checker enabled
- `npx nx build @midscene/android` — passed (downstream consumer of the widened
  `BaseAgent.aiAction` type)
- pre-commit `lint-staged` (biome) — passed on the changed files

### DIFF
--- a/packages/shared/src/agent-tools/tool-generator.ts
+++ b/packages/shared/src/agent-tools/tool-generator.ts
@@ -712,6 +712,7 @@ export function generateCommonTools(
           .describe(
             'Plan this action with deep thinking (richer context and sub-goal decomposition). Helps with complex multi-step instructions at the cost of speed. Defaults to the server --deep-think setting.',
           ),
+        ...promptInputExtraSchema,
         ...initArgSchema,
       },
       cli: mergeToolCliMetadata(undefined, initArgCliMetadata),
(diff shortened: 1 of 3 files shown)


codeestX:
Superseded by an origin-branch PR so upstream CI runs (fork PRs stall). Reopening the same change from web-infra-dev/midscene:feat/cli-act-image-prompt.

PR #2958, opened 2026-08-11
chore(workflow): migrate shared, computer, android and ios to rstest

fi3ework:
Third batch of the vitest 3.0.5 -> rstest 0.11.5 migration, after #2921 and #2945: `shared`, `computer`, `android`, `ios`. 118 test files. Remaining after this: `cli` + `web-integration`, then `core`, then `studio` plus the vitest teardown.
All but five files are a pure `vitest` -> `@rstest/core` specifier swap plus `vi.` -> `rs.`. Those five are under "What actually needs eyes".

Test discovery is unchanged: **80 files / 1089 cases before and after** on the default unit include, and not just the totals — I diffed the full `file > test name` list from `rstest list` against `vitest list` on a worktree of the pre-migration tree, across all nine include modes these packages have (unit plus each `AITEST` / `AI_TEST_TYPE`). Every mode is identical. The AI suites need real devices and models so they are not executed here, but `rstest list` compiles them, which is what caught two of the type problems below.

## Performance

Roughly break-even overall, with the win concentrated in `shared`. Median of 3 runs per cell, runners alternated each round so machine drift hits both equally; cold clears the bundler caches first. Apple M3 Max, Node 26.5.0, both runners exit 0 on every sample.

| Package | Cold vitest | Cold rstest | | Warm vitest | Warm rstest | |
| :-- | --: | --: | --: | --: | --: | --: |
| shared | 2.96s | 1.66s | **1.78x** | 2.88s | 1.41s | **2.04x** |
| computer | 11.92s | 10.73s | 1.11x | 11.60s | 10.96s | 1.06x |
| android | 13.44s | 12.70s | 1.06x | 13.45s | 12.73s | 1.06x |
| ios | 18.04s | 18.30s | 0.99x | 17.83s | 18.14s | 0.98x |
| **total** | **46.36s** | **43.39s** | **1.07x** | **45.76s** | **43.24s** | **1.06x** |

The spread follows the same rule the `chrome-extension` result in #2945 did: **the gain tracks the compile-to-execute ratio, not the file count.** rstest's advantage is compilation, so a suite that barely compiles and mostly waits has nothing to win. rstest reports the split itself:

| Package | build | tests | build share |
| :-- | --: | --: | --: |
| shared | 71ms | 886ms | 7.4% |
| computer | 219ms | 10.1s | 2.1% |
| android | 275ms | 12.8s | 2.1% |
| ios | 229ms | 17.2s | 1.3% |

`shared` is 441 fast pure-function tests where compilation is the dominant movable cost, and it doubles warm. `ios` spends 1.3% of its run building, which caps any bundler-side gain at about 1.3% — inside the noise. Measured file by file, `device.test.ts` alone takes 17.38s wall while the whole package takes 17.4s, so it is the critical path; inside it the 48 cases account for only 2.26s and the remaining ~15s is `afterEach` calling `await device.destroy()` 48 times against a `src/device.ts` that sleeps for real (`sleep(100)`, `sleep(500)`, `sleep(800)`, `sleep(2000)`). `computer` and `android` are the same shape.

These four are here for uniformity, not speed; the batches that moved the needle were the cold ones in #2945. Unit tests only — `tests/ai/**` is outside the default include, so no model or network calls are involved.

## What actually needs eyes

**Async mock factories have no rstest equivalent** (android, 3 sites). rstest's `MockFactory` is synchronous, so `vi.mock('x', async (importOriginal) => ...)` is a hard error. `page.test.ts` (x2) and `scrcpy-adapter.test.ts` move to the import-attribute recipe `harmony` already uses in #2945:

```ts
import * as fsActual from 'node:fs' with { rstest: 'importActual' };
rs.mock('node:fs', () => { const original = fsActual; /* ... */ });
```

**`ios/tests/unit-test/agent.test.ts` stops mocking the device-class override.** Two blockers stack: `agentFromWebDriverAgent` resolves the override through `await import(overrideModule)` where the specifier is a user-supplied runtime value, which rstest's build-time mock transform cannot reach (web-infra-dev/rstest#1454); and rstest has no virtual-module mocking, its `MockModuleOptions` being only `{ spy: true } | { mock: true }`. The two cases now point at real `.mjs` fixtures and observe them through a global counter. Nothing is skipped, and the suite gains an assertion that the default device class stays unused, which the mock-based version could not express.

Worth knowing: the old code cast `vi.doMock` to a signature vitest does not have — `{ virtual: true }` is a Jest option, vitest's `MockOptions` is `{ spy?: boolean }`. Dropping that argument on the pre-migration tree leaves all 13 cases green, so it was decorative, and it meant the two success-path cases never exercised real module resolution while the failure-path case did.

**A dead flag, removed rather than translated.** `android`, `ios` and `harmony` carried `errors: process.env.CI ? { unhandled: false } : undefined`, translated from `dangerouslyIgnoreUnhandledErrors`. It never did that job: in rstest 0.11.5 `errors.unhandled` belongs to `MdReporterOptions`, not the top-level config, so at most it decides whether the markdown reporter prints an error section. A passing test that leaks a rejection exits 1 with the field and without it.

This PR deletes it instead of looking for a working equivalent. The flag was inherited rather than earned — its comment blames `showcase.test.ts`, a path that has never existed anywhere in this repository's history, and the teardowns it was meant to cover already await their work inside `try`/`catch`. Suppressing unhandled rejections hides exactly the failures worth seeing; if CI surfaces one, the leak is the thing to fix. `harmony` is included because the same line landed with #2945, inert there too.

**Three type errors nothing was checking.** The AI suites are compiled but not type-checked by lint or `nx test`, so these only surface under `tsc`:

- `android/tests/ai/merge-reports.test.ts` read `ctx.task.result.state` and `.startTime`; rstest names the field `status` and carries no start timestamp, so the duration now uses the `performance.now()` stamp the suite already takes.
- `computer/tests/ai/chrome-extension-bridge.test.ts` passed `{ timeout: 20 * 60 * 1000, retry: 0 }` as the third argument. rstest takes options in the second position; in the third it is **silently ignored**, so that suite would have quietly fallen back to the config's 3-minute `testTimeout`.
- `android/tsconfig.json` pinned `module: ES2020`, which rejects the import-attribute syntax. The pin moves to `tsconfig.build.json` so the production build keeps its module semantics while tests inherit `ESNext`. A stale `tsconfig.tsbuildinfo` masks this one if you reproduce locally.

**Nothing type-checked the tooling.** The misplaced `errors` key survived because `scripts/**` and every `rstest.config.ts` sit outside all package tsconfig `include` lists. `tsconfig.tools.json` now covers both and runs first in `pnpm run type-check:tests`; putting the bad key back produces `TS2769` at that line.

## Config translation

`test.*` moves to the top level, `ssr.external` -> `output.externals`, `define` -> `source.define`, `fileParallelism: false` -> `pool: { maxWorkers: 1 }` (computer, android). `computer`'s explicit `environment: 'node'` is kept as `testEnvironment: 'node'` rather than dropped as redundant.

`scripts/rstest-shared.ts` is new: the photon external and the `__VERSION__` define were repeated verbatim in every node-target config, so with five consumers they are defined once and `harmony` moves over too. Deferred from #2945, where `harmony` was the only consumer.

Also here, the follow-ups from the #2945 review: `harmony/tsconfig.build.json` pins `module: ES2020` again so the production build is decoupled from test-only syntax, two stale `@ts-ignore` comments are gone, and `scripts/rstest-style-stub.ts` types its rspack callback as `Rspack.Configuration` — the suggested `satisfies Pick<RstestConfig, 'tools'>` alone catches nothing there, because `tools.rspack` is a `ConfigChain` union.

### DIFF
--- a/packages/android/package.json
+++ b/packages/android/package.json
@@ -36,10 +36,10 @@
     "build:watch": "rslib build --watch --no-clean",
     "prepack": "node scripts/download-scrcpy-server.mjs && node scripts/download-yadb.mjs",
     "playground": "DEBUG=midscene:* tsx demo/playground.ts",
-    "test": "vitest --run",
-    "test:u": "vitest --run -u",
-    "test:ai": "AI_TEST_TYPE=android vitest --run",
-    "test:ai:cache": "MIDSCENE_CACHE=true AI_TEST_TYPE=android vitest --run"
+    "test": "rstest",
+    "test:u": "rstest -u",
+    "test:ai": "AI_TEST_TYPE=android rstest",
+    "test:ai:cache": "MIDSCENE_CACHE=true AI_TEST_TYPE=android rstest"
   },
   "dependencies": {
     "@midscene/core": "workspace:*",
@@ -58,12 +58,12 @@
   "devDependencies": {
     "@midscene/playground": "workspace:*",
     "@rslib/core": "^0.18.3",
+    "@rstest/core": "0.11.5",
     "@types/node": "^18.0.0",
     "dotenv": "^16.4.5",
-    "typescript": "^5.8.3",
     "tsx": "^4.19.2",
+    "typescript": "^5.8.3",
     "undici": "^6.0.0",
-    "vitest": "3.0.5",
     "zod": "^3.25.1"
   },
   "license": "MIT"
--- a/packages/android/rstest.config.ts
+++ b/packages/android/rstest.config.ts
@@ -35,7 +35,6 @@ export default defineConfig({
   coverage: createCoverageConfig(__dirname),
   include: testFiles,
   testTimeout: 3 * 60 * 1000, // Global timeout set to 10 seconds
-  errors: process.env.CI ? { unhandled: false } : undefined, // showcase.test.ts is not stable
   pool: { maxWorkers: 1 }, // disable parallel file test
   source: {
     define: defineVersion(version),
--- a/packages/android/rstest.config.ts
+++ b/packages/android/rstest.config.ts
@@ -1,7 +1,8 @@
 import path from 'node:path';
+import { defineConfig } from '@rstest/core';
 import dotenv from 'dotenv';
-import { defineConfig } from 'vitest/config';
-import { createCoverageConfig } from '../../scripts/vitest-coverage';
+import { createCoverageConfig } from '../../scripts/rstest-coverage';
+import { defineVersion, photonExternal } from '../../scripts/rstest-shared';
 import { version } from './package.json';
 
 /**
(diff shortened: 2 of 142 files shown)


fi3ework:
Superseded by #2999 — same branch pushed to the upstream repo directly so CI can read secrets (fork PRs cannot).

PR #2992, opened 2026-08-17
refactor(core): isolate report template generation

EAGzzyCSL:
## Summary

- isolate the embedded Report HTML in fixed CJS/ESM template modules
- share one template validation/generation path between Core and Report builds
- replace CI/E2E fallback injection with read-only validation and exclude runtime template modules from Nx cache
- remove Studio's redundant Core rebuild fallback and validate the generated template modules instead

## Why

Core and Report previously scanned and rewrote arbitrary Core bundles, with multiple injection paths and cache-sensitive side effects. Keeping the report in a dedicated module makes the coupling explicit and keeps local Core-only builds convenient while Report builds remain the authoritative update point.

## Behavior

- Core builds reuse an existing valid Report, or keep the placeholder intermediate module
- Report builds use `IS_REPORT_BUILD` to avoid recursive embedding and overwrite both template modules
- release, CI, and Studio packaging validation fail on missing/invalid templates without repairing them

## Validation

- `pnpm exec biome check . --diagnostic-level=info --no-errors-on-unmatched`
- `pnpm exec rstest run tests/report-template-utils.test.ts` (3 tests)
- Studio packaging tests (65 tests)
- focused builds for shared, core, playground, web-integration, visualizer, and report
- repeated Report builds remained size-stable
- CJS/ESM `getReportTpl()` smoke checks
- `pnpm pack --dry-run --json`


mamico:
U

PR #3007, opened 2026-08-20
chore(workflow): migrate cli and web-integration to rstest

fi3ework:
Fourth batch of the vitest 3.0.5 -> rstest 0.11.5 migration, after #2921, #2945 and #2999: `cli` + `web-integration`. 105 test files; all but a handful are the mechanical `vitest` -> `@rstest/core` specifier swap plus `vi.` -> `rs.`. Remaining after this: `core`, then `studio` plus the vitest teardown.

Test discovery is unchanged in every include mode — CLI unit **16 files / 203 cases**, CLI `AITEST` **1 / 7**, web unit **51 / 431 (+1 pre-existing skip)**, web `AI_TEST_TYPE=web` — verified by diffing the full `file > test name` list from `rstest list` against `vitest list` on the pre-migration tree, case by case, not just totals. Zero lost cases, zero new skips. The AI suites need a real model and browser so they are not executed here, but `rstest list` compiles them, which is what surfaced the discovery crash described below.

## Performance

Median of 3 warm runs after one cold run per cell; cold clears the bundler caches first. Apple M3 Max, Node 26.5.0.

| Package | Cold vitest | Cold rstest | | Warm vitest | Warm rstest | |
| :-- | --: | --: | --: | --: | --: | --: |
| cli | 8.16s | 4.81s | **1.70x** | 5.46s | 3.82s | **1.43x** |
| web-integration | 26.09s | 24.94s | 1.05x | 25.57s | 25.18s | 1.02x |
| **total** | **34.24s** | **29.75s** | **1.15x** | **31.02s** | **28.99s** | **1.07x** |

Same rule as #2999: the gain tracks the compile-to-execute ratio. `cli` is transform-heavy with fast tests and gets the full win. `web-integration` spends its time in real browsers — puppeteer page launches, bridge-io servers, screenshots, and a circular-dependency suite that imports the built package (4.6s on its own); rstest builds the suite in ~0.15s versus vitest's ~1.4s transform, which is noise against a 25s browser-bound run.

## What actually needs eyes

**Three AI test files called `getModelConfig('default')` at module top level, which throws when no model env is configured** (`ai-number`, `search`, `ui-observer`). Under both runners this crashed collection on an env-less machine — the `skipIf(!canRunAiTest)` these files carry never got the chance to skip anything. They now guard on `process.env.MIDSCENE_MODEL_NAME` before resolving the config, so an unconfigured checkout skips instead of erroring. CI behavior is unchanged: the env is always set there, and the resolved config is identical.

**The original `dangerouslyIgnoreUnhandledErrors` is gone** (`web-integration/vitest.config.ts:41`). This is the line whose comment blames `showcase.test.ts` — a file that has never existed in the repository — and whose translated copies #2999 already removed from `android`/`ios`/`harmony`. rstest has no equivalent, and the flag's job (turning a run that leaks unhandled rejections green in CI) is not one worth doing. Deleted; the full unit suite leaks nothing.

**vitest accepted an options object in the third argument; rstest silently drops it there.** Affected sites moved to the second position, same as the `chrome-extension-bridge` fix in #2999. Also `TestResult.state` -> `.status` and local timing instead of the removed `startTime`, in the playwright reporter tests.

**Ten comment-only lines in `src/puppeteer/base-page.ts`** — the one production file touched, found by the first CI round. The coverage run instruments source with istanbul (v8 OOMs under `pool.maxWorkers: 1`, see the note in `scripts/rstest-coverage.ts`), and closures handed to `page.evaluate` are serialized into the browser realm where the injected `cov_*` counter does not exist, so every such call crashed with a `ReferenceError` once coverage was on. Each call site now carries `/* istanbul ignore next */`; annotating the sites instead of excluding the file keeps the rest of `base-page.ts` in the report, and the full web suite is green under `--coverage` locally.

**Async mock factories** (9 files across both packages) move to the same import-attribute recipe as #2945/#2999: `import * as XActual from 'x' with { rstest: 'importActual' }` plus a synchronous factory.

Small mechanical rest: `static-page.test.ts` was skipped by the codemod because of its `static/` path and is migrated by hand (one import line); 7 snapshot files change only their header (`// Vitest Snapshot v1` -> `// Rstest Snapshot v1`).

## Config translation

`test.*` moves to the top level, `ssr.external` -> `output.externals`, `define` -> `source.define`, CI `minWorkers/maxWorkers` -> `pool: { maxWorkers: 4 }`, and the CI-only `outputFile: { json: './test-results.json' }` becomes the `['json', { outputPath: './test-results.json' }]` reporter, which is rstest's way of writing the same file. Both packages reuse `scripts/rstest-shared.ts` and `createCoverageConfig` from the previous batches.


fi3ework:
@quanru @EAGzzyCSL PTAL~

quanru (approved):

PR #3019, opened 2026-08-21
fix(core): deduplicate multimodal reference images in reports

quanru:
## Summary

- persist base64 multimodal prompt images as content-addressed report assets instead of repeating the payload in every dump
- register prompt images explicitly when an execution session is created; `ExecutionDump` now carries non-serialized sidecar metadata and no longer scans arbitrary task parameters
- serialize screenshots and prompt images with distinct typed references owned by `ReportImageStore`
- restore both stored-image reference types through one `StoredImageRef` resolver and one MIME-aware path resolver
- support PNG, JPEG, WebP, SVG, GIF, AVIF, BMP, and safe image subtypes in inline and directory reports
- deduplicate inline image tags across replanning tasks, generator restarts, incremental updates, and streaming report merges
- keep Page, Browser, custom-serializable values, and other class instances opaque during dump serialization
- restore visible progress connectors in the dark Chrome Extension timeline, including multi-line rows
- keep the Chrome Extension clear control inside the panel boundary
- stabilize Action mode selection when the mode is hidden in the horizontal selector
- stabilize report replay coverage without replacing its core AI assertions with DOM-only checks

## Root cause

Dump compaction removed superseded dump tags, but base64 reference images could still grow a report through three independent paths:

1. the retained execution serialized the same reference image in every replanning task;
2. a new `ReportGenerator` reused an existing report without knowing which inline image IDs had already been written;
3. `ReportMergingTool` copied matching inline image tags from every source report.

The fix gives persisted report images one owner, replaces registered base64 URLs with content-addressed references at serialization time, restores those references at consumer boundaries, and shares written-image state across reuse and merge flows.

## Review follow-ups

- removed the synchronous whole-report scan from `ReportImageStore` construction
- removed the unbounded cache of full data URLs
- replaced late recursive `collectReferenceImageUrls` inference with explicit registration at `aiAct`, YAML, Query/Assert, and WaitFor execution boundaries
- kept the registration sidecar out of the report JSON schema
- consolidated `restoreImageReferences` from two equivalent callbacks to one typed `StoredImageReferenceResolver`
- removed JPEG-vs-PNG fallback logic from report and visualizer replay consumers
- replaced object-identity coupling with occurrence-scoped serialization and URL-to-asset lookup
- moved directory and inline data-URI loading into `ReportImageStore`
- routed report extractor dependencies through the public `@midscene/core/dump` entry
- preserved opaque runtime object handling without traversing their getters
- split the report-generator regression suite into focused files below the repository file-size limit
- made the shared visualizer own timeline connector geometry and let the Chrome Extension provide only its theme color
- removed the duplicate extension connector that obscured the shared high-contrast line
- reused the focused Action-mode prompt from #3027 because that PR was closed without merging

## Regression coverage

- compare a single-task report with a 20-task report and reject growth by another full image payload
- compare 20 incremental execution snapshots with a one-shot final snapshot and reject retained history
- verify 50-task directory and split reports contain one external image and bounded reference-only JSON
- reuse the same inline report from two generator instances and require one image payload
- merge reports that share an inline reference image and require one emitted image tag
- verify WebP replay URLs, SVG persistence/restoration, and MIME-derived extensions
- ensure a reused descriptor does not rewrite an unrelated field
- verify task parameters are not recursively inspected for reference-image registration
- verify `aiAct`, Insight, and TaskRunner propagate registered image metadata without serializing it
- verify the unified resolver receives both `ScreenshotRef` and `ImageUrlRef`
- verify non-final timeline rows receive connectors and dark mode uses a visible connector color
- verify the Chrome Extension does not reintroduce a second connector implementation

## Validation

- `pnpm run lint`
- `pnpm exec nx test @midscene/core` (1481 passed, 8 skipped)
- `pnpm exec nx test @midscene/playground` (273 passed)
- `pnpm exec nx test @midscene/visualizer` (115 passed)
- `pnpm exec nx test @midscene/report` (97 passed)
- `pnpm exec nx test chrome-extension` (54 passed)
- `pnpm exec nx run-many --target=build --projects="@midscene/core,@midscene/playground,@midscene/visualizer,@midscene/report,chrome-extension"`
- `pnpm exec nx run-many --target=build --projects="@midscene/report,chrome-extension" --skip-nx-cache`
- `pnpm exec tsc -p packages/core/tsconfig.json --noEmit`
- `pnpm exec tsc -p packages/visualizer/tsconfig.json --noEmit`
- `pnpm exec tsc -p apps/report/tsconfig.json --noEmit`
- `git diff --check`

## Prepatch

- previous build: `@midscene/core@1.11.1-beta-20260824074440.0` from `8836d4182`
- latest published build: `@midscene/core@1.11.1-beta-20260825110930.0` (`beta` dist-tag) from `c6d78899f`; [release workflow](https://github.com/web-infra-dev/midscene/actions/runs/32840707739)


quanruzhuoxiu:
Technical guide (Chinese): https://bytedance.larkoffice.com/docx/PWzOd0boIo83LPx2ofNcPLkcnrc\n\nIt documents the before/after size evidence, root causes, final ownership boundaries, review follow-ups, regression coverage, prepatch version, and deferred non-blocking optimizations.

EAGzzyCSL (approved):

PR #3066, opened 2026-08-28
fix(core): commit planning logs after actions succeed

linhongyu510:
## Summary

- defer planning-log persistence until the corresponding action batch succeeds
- prevent failed actions from appearing as completed steps in later planning rounds
- preserve normal and deepThink history semantics
- skip log persistence for empty action batches
- add regression coverage for successful, failed, empty, and deepThink executions

## Problem

`standardPlan()` previously appended the model's planning log to `ConversationHistory` before the planned action was executed.

If the action later failed, the next planning round received contradictory context: the execution history said the action had already been performed, while pending feedback reported that execution failed. This could cause the GUI Agent to skip retries, advance to the wrong sub-goal, or make decisions from an incorrect execution state.

## Changes

Planning-log ownership now lives in `TaskExecutor`, which knows whether the action batch actually completed.

- normal effort: append successful logs to historical execution steps
- deepThink effort: append successful logs to the running sub-goal
- failed batch: do not append
- empty batch: do not append

Sub-goal planning updates and model conversation messages remain in `standardPlan()` because they describe planning state rather than confirmed action execution.

## Validation

- planning-history regression suite: 4 tests passed
- focused planning suite: 15 tests passed
- Biome: 1605 files passed

Covered cases:

- successful action commits its log exactly once
- failed action does not commit its log
- empty action batch does not commit its log
- deepThink mode commits the log to the running sub-goal

### DIFF
--- a/packages/core/src/agent/tasks.ts
+++ b/packages/core/src/agent/tasks.ts
@@ -251,6 +251,27 @@ export class TaskExecutor {
   }
 
   /**
+   * Commit the model's action log only after the corresponding action batch
+   * succeeds. Planning logs describe work to be performed, so recording them
+   * before execution would make failed actions appear completed on the next
+   * planning round.
+   */
+  private commitSuccessfulPlanningLog(
+    conversationHistory: ConversationHistory,
+    log: string | undefined,
+    effort: AiActEffort,
+  ): void {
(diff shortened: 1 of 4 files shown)


linhongyu510:

PR #3067, opened 2026-08-28
fix(core): reject negative DeepSeek coordinates

linhongyu510:
## Summary

- preserve coordinate signs when parsing DeepSeek grounding responses
- reject negative point coordinates instead of silently converting them to positive values
- reject negative bbox coordinates in every position
- add regression coverage while preserving supported coordinate separators

## Problem

The DeepSeek coordinate parser previously extracted values with `/\d+/g`. As a result, a model response such as `[-10, 500]` was silently interpreted as `[10, 500]`, allowing malformed grounding output to become a valid click coordinate.

The same issue affected bbox responses: a negative edge could lose its sign before the shared coordinate-range validation saw it.

## Changes

The parser now preserves an optional sign while extracting integer values and rejects any negative coordinate before returning normalized values. Existing support for bracket variants and flexible separators is unchanged.

Regression tests cover negative values in both point positions and all four bbox positions.

## Validation

- DeepSeek adapter tests: 20 passed
- Biome: 1604 files passed

### DIFF
--- a/packages/core/src/ai-model/models/deepseek/adapter.ts
+++ b/packages/core/src/ai-model/models/deepseek/adapter.ts
@@ -30,14 +30,18 @@ function parseDeepSeekCoordinateValues(
   expectedLength: number,
   label: string,
 ): number[] {
-  const coordinateTexts = String(input).match(/\d+/g);
-  if (coordinateTexts?.length !== expectedLength) {
+  const coordinateTexts = String(input).match(/[+-]?\d+/g);
+  const coordinates = coordinateTexts?.map(Number);
+  if (
+    coordinates?.length !== expectedLength ||
+    !coordinates.every((coordinate) => coordinate >= 0)
+  ) {
     throw new Error(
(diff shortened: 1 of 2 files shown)


linhongyu510:

linhongyu510:
Follow-up on the CI, same fingerprint as I noted before.

The red checks here are the model-backed / device jobs (`e2e-web`, `chrome-extension-*`, `android-emulator`, `ios-simulator`, `macos/windows-desktop`) that depend on the injected secrets on an external-fork run — not the changed code. This PR's change is confined to `packages/core` (DeepSeek adapter negative-coordinate parsing) with its own unit coverage in `deepseek.test.ts`; there is no `core` unit-test job among the failures, they are all the environment-backed matrix.

I verified the same class of change on the sibling PR #3068 by running its `packages/core` unit file locally (33/33 passing), so the parsing logic itself is sound. A maintainer can validate via a trusted branch or `workflow_dispatch`; happy to rebase if that helps.

PR #3068, opened 2026-08-29
fix(core): validate locate coordinate types strictly

linhongyu510:
## Summary

- reject null, boolean, empty, and whitespace-only coordinate values
- preserve support for valid decimal coordinate strings
- require point results to contain exactly two coordinates
- add regression coverage for JavaScript coercion edge cases

## Problem

The shared locate-result parser used `Number(value)` for arbitrary values. This allowed malformed model output such as `null`, `false`, or an empty string to become finite coordinates and proceed into the grounding pipeline:

```ts
Number(null)  // 0
Number(false) // 0
Number(true)  // 1
Number('')    // 0
```

Point responses with extra coordinates were also silently truncated because the parser only required at least two values.

## Changes

Coordinate values now accept only finite numbers or complete decimal number strings. General JavaScript coercion is no longer used, and point responses must contain exactly two values.

This keeps compatibility with numeric strings while ensuring malformed model output reaches the existing semantic-retry path instead of becoming a real device coordinate.

## Validation

- locate-result codec tests: 33 passed
- model-adapter and codec suite: 28 files, 314 tests passed
- Biome: 1604 files passed

### DIFF
--- a/packages/core/src/ai-model/shared/model-locate-result/parse.ts
+++ b/packages/core/src/ai-model/shared/model-locate-result/parse.ts
@@ -43,9 +43,18 @@ export function parseCoordinateList(input: unknown, label: string): number[] {
     throw new Error(`invalid ${label} data: ${JSON.stringify(input)} `);
   }
 
-  const numericValues = values.map((value) =>
-    typeof value === 'number' ? value : Number(value),
-  );
+  const numericValues = values.map((value) => {
+    if (typeof value === 'number') {
+      return value;
+    }
+    if (
+      typeof value !== 'string' ||
+      !/^[+-]?(?:\d+\.?\d*|\.\d+)$/.test(value.trim())
+    ) {
+      return Number.NaN;
+    }
+    return Number(value);
+  });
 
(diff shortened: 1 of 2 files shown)


linhongyu510:

linhongyu510:
Follow-up with local evidence for the changed code itself.

I cloned this branch, installed `packages/core`, and ran the unit file this PR adds/changes:

```
packages/core $ npx vitest --run tests/unit-test/locate-result-codec.test.ts
 ✓ tests/unit-test/locate-result-codec.test.ts (33 tests)
 Test Files  1 passed (1)
      Tests  33 passed (33)
```

So the coordinate-parsing change (`parseCoordinateList` rejecting values that only `Number()`-coerce — `null`/`true`/`''`/whitespace — and the strict `point.length !== 2`) is green locally. The red checks on the PR are the model-backed / device jobs (`e2e-web`, `chrome-extension-*`, `android-emulator`, `ios-simulator`, `macos/windows-desktop`) that need the injected secrets, which is the same external-fork gate I described above — not the changed logic. Happy to rebase if it helps a trusted rerun.

PR #3069, opened 2026-08-29
fix(core): validate swipe action parameters

linhongyu510:
## Summary

- require swipe distance and duration to be positive finite numbers
- require repeat to be a non-negative finite integer
- reject end and distance when supplied together
- require direction for distance-based swipes
- preserve the existing repeat=0 safety cap and clarify its description
- add focused schema and runtime regression coverage

## Problem

Invalid Swipe parameters could previously produce unexpected device actions:

- negative distance reversed the requested swipe direction
- negative repeat silently executed zero times
- fractional repeat could execute more times than requested
- invalid duration values reached platform adapters
- distance was silently ignored when end was also present

The repeat=0 description also claimed infinite execution, while the implementation intentionally capped it at 10 iterations.

## Changes

Validation is applied at two boundaries:

1. the action schema validates model-generated scalar parameters
2. `normalizeMobileSwipeParam()` validates runtime and direct SDK input

Cross-field constraints remain in the runtime normalization layer so the schema stays a ZodObject and existing locator-field extraction continues to work.

## Validation

- Swipe regression suite: 20 tests passed
- device and related planning suite: 6 files, 73 tests passed
- Biome: 1605 files passed

### DIFF
--- a/packages/core/src/device/index.ts
+++ b/packages/core/src/device/index.ts
@@ -760,6 +760,8 @@ export const ActionSwipeParamSchema = z.object({
     ),
   distance: z
     .number()
+    .finite()
+    .positive()
     .optional()
     .describe('The distance in pixels to swipe (mutually exclusive with end)'),
   end: getMidsceneLocationSchema()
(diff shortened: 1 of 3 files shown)


linhongyu510:
Note on the failing checks, in case it factors into review.

This PR's change is confined to `packages/core` (swipe-parameter validation) with unit coverage in `swipe.test.ts` and `action-description.test.ts`. The red checks are all the model-backed / device jobs (`e2e-web`, `chrome-extension-*`, `android-emulator`, `ios-simulator`, `macos/windows-desktop`) that need the repository secrets injected on the run — the same external-fork trust/configuration gate, not the changed logic. There is no `core` unit-test job among the failures.

I confirmed this class of change builds and passes locally on the sibling PR #3068 (`packages/core` unit file, 33/33). A maintainer can validate via a trusted branch or `workflow_dispatch` after reviewing the diff; happy to rebase if it helps.

PR #3071, opened 2026-08-31
feat(core): support glm-5.3-flash under glm-v family

kinsolee:
# feat(core): support glm-5.3-flash under glm-v family

## Summary

GLM-5.3-Flash is an **always-thinking** vision model: it rejects
`thinking.type: 'disabled'` with error `1210` ("该模型始终思考，不支持关闭思考"),
so the existing glm-v adapter contract does not fit it.

This PR branches the glm-v chat-completion parameter builder on the resolved
model name:

- For `glm-5.3-flash` (matched by `/^glm-5\.3-flash\b/`):
  - thinking stays **enabled** with `clear_thinking: false`
  - depth is steered via `reasoning_effort` (defaults to `low` when the user
    disables reasoning, since thinking cannot be turned off)
  - officially recommended sampling defaults: `temperature: 1`, `top_p: 0.95`
    (https://docs.bigmodel.cn/cn/guide/models/vlm/glm-5.3-flash)
- All other glm-v models keep the existing toggleable-thinking contract
  unchanged.

To support the branching, `ChatCompletionCallInput` /
`ChatCompletionCallContext` gain an optional `modelName`, and `callAI` passes
`modelConfig.modelName` through.

## Changes

- `packages/core/src/ai-model/models/glm.ts` — always-thinking branch
- `packages/core/src/ai-model/model-adapter/types.ts` — optional `modelName`
- `packages/core/src/ai-model/service-caller/index.ts` — pass `modelName`
- `packages/core/tests/unit-test/model-adapter/glm.test.ts` — new unit tests
- `apps/site/docs/{en,zh}/model-common-config.mdx` — list glm-5.3-flash under
  glm-v with an always-thinking note

## Verification

Field-tested against the Zhipu (BigModel) Coding Plan endpoint on 2026-08-26
(KSL-49 spike) and re-verified on 2026-08-31 in a production Android-automation
setup driving real devices through `agentFromAdbDevice`:

- `glm-5.3-flash` with the patched parameters returns HTTP 200 and completes
  real `aiAct` tasks on physical Android devices (e.g. "open the Settings app"
  reliably launches `com.android.settings`).
- Same-sample accuracy on par with the production vision models
  (qwen3.8-max / MiniMax-M3); locate IoU 0.698 with the glm-v 0-1000 bbox
  contract on a UI-screenshot benchmark.
- `thinking.type: 'disabled'` is confirmed rejected by the endpoint with error
  1210, which is why the adapter keeps thinking enabled for this model.
- The added unit tests cover both the always-thinking branch and the unchanged
  toggleable-thinking behavior of other glm-v models.

### DIFF
--- a/apps/site/docs/en/model-common-config.mdx
+++ b/apps/site/docs/en/model-common-config.mdx
@@ -400,6 +400,7 @@ MIDSCENE_INSIGHT_MODEL_FAMILY="xiaomi-mimo"
 
 | Model version | Commonly used model names | `MIDSCENE_MODEL_FAMILY` | Notes |
 | --- | --- | --- | --- |
+| GLM-5.3 series | `glm-5.3-flash` | `glm-v` | Always-thinking: `thinking.type` cannot be disabled, so disabling reasoning sends `reasoning_effort: low` instead. |
 | GLM-5V series | `glm-5v-turbo` | `glm-v` | — |
 | GLM-4.6 series | `glm-4.6v` | `glm-v` | `glm-4.6v` is open-source. |
 
--- a/apps/site/docs/en/model-config.mdx
+++ b/apps/site/docs/en/model-config.mdx
@@ -107,6 +107,7 @@ The following model families currently support `MIDSCENE_MODEL_REASONING_ENABLED
 - Gemini: Maps to `thinking_config.thinking_level`.
 - GPT-5: Maps to `reasoning_effort`.
 - Kimi K3 series: Maps to `reasoning_effort`.
+- Zhipu GLM: Maps to `reasoning_effort` (glm-5.3-flash only).
 
 Provider support and accepted values vary. See each provider's official documentation for supported model versions and values. If the current model does not support an explicit setting, Midscene ignores it instead of guessing a provider-specific private parameter.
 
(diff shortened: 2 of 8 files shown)


kinsolee:
Thanks for the review — all three points are addressed in 19be45cb:

1. **Effort override in default mode (glm.ts:59)**: `reasoningEnabled='default'` now sends neither a thinking nor an effort override; explicit `MIDSCENE_MODEL_REASONING_EFFORT` is ignored in that mode per the documented contract.
2. **Forced `low` when reasoning is enabled (glm.ts:57)**: the `low` mapping now applies only to the no-reasoning intent (explicit `false` or unset). Explicitly enabling reasoning without an effort keeps the provider default; an explicit effort is still forwarded.
3. **Docs (model-config.mdx)**: Zhipu GLM (glm-5.3-flash only) added to the `MIDSCENE_MODEL_REASONING_EFFORT` supported-families list in both en and zh.

Both behavior changes are covered by new unit tests (`keeps the provider effort default when reasoning is explicitly enabled without effort`, `ignores an explicit effort in default reasoning mode for glm-5.3-flash`).

@codex review

PR #3094, opened 2026-09-03
fix(core): preserve global model config with custom clients

### LINKED ISSUE #3079: [Bug]: `createOpenAIClient` option breaks model configuration
### Version

```sh
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 8.65 GB / 31.84 GB
  Browsers:
    Chrome: 151.0.7922.138
    Edge: Chromium (152.0.4191.53)
    Internet Explorer: 11.0.26100.7309
  npmPackages:
    @midscene/cli: ^1.12.2 => 1.12.2 
    @midscene/core: 1.12.2 => 1.12.2 
    @midscene/web: ^1.12.2 => 1.12.2
```

### Details

# Bug Report: `createOpenAIClient` option breaks model configuration

## Environment

- **OS**: Windows 11
- **Package**: @midscene/core@1.12.2, @midscene/web@1.12.2, @midscene/shared@1.12.2
- **Node.js**: 20+
- **Test Framework**: Vitest with e2e project

## Summary

When passing `createOpenAIClient` option to `PlaywrightAgent` (or other Midscene agents), the test fails with `Model configuration is incomplete: model name (MIDSCENE_MODEL_NAME) is required` error, even though environment variables are correctly set.

## Steps to Reproduce

```typescript
import { PlaywrightAgent } from '@midscene/web/playwright';

const agent = new PlaywrightAgent(page, {
  groupName: 'test',
  reportFileName: 'test-report',
  cache: { id: 'test-cache' },
  // Even an empty/no-op hook breaks the configuration
  createOpenAIClient: async (baseOpenAI) => {
    return baseOpenAI;
  },
});
```

## Expected Behavior

`createOpenAIClient` should be called with the base OpenAI client, allowing users to wrap it with observability tools like Langfuse, without affecting model configuration.

## Actual Behavior

Tests fail immediately with:
```
Error: Model configuration is incomplete: model name (MIDSCENE_MODEL_NAME) is required. See https://midscenejs.com/model-strategy.html
```

## Root Cause Analysis

In `node_modules/@midscene/core/dist/es/agent/agent.mjs` lines 815-816:

```javascript
const hasCustomConfig = opts?.modelConfig || opts?.createOpenAIClient;
this.modelConfigManager = hasCustomConfig ? new ModelConfigManager(opts?.modelConfig, opts?.createOpenAIClient) : globalModelConfigManager;
```

When `createOpenAIClient` is provided (even without `modelConfig`), a **new** `ModelConfigManager` instance is created instead of using the global singleton.

However, this new instance is not properly initialized:

1. `globalModelConfigManager` (the singleton) is created in `utils.mjs` with:
   ```javascript
   const globalModelConfigManager = new ModelConfigManager();
   const globalConfigManager = new GlobalConfigManager();
   globalConfigManager.registerModelConfigManager(globalModelConfigManager);
   globalModelConfigManager.registerGlobalConfigManager(globalConfigManager);  // ← registers globalConfigManager
   ```

2. The new instance created when `createOpenAIClient` is passed:
   ```javascript
   new ModelConfigManager(opts?.modelConfig, opts?.createOpenAIClient)
   ```
   - Only passes `(modelConfig, createOpenAIClientFn)`
   - Does **NOT** call `registerGlobalConfigManager()`
   - Therefore `this.globalConfigManager` is undefined

3. In `ModelConfigManager.initialize()`:
   ```javascript
   if (this.modelConfig) {
     this.isolatedMode = true;
     configMap = this.normalizeModelConfig(this.modelConfig);
   } else configMap = this.globalConfigManager?.getAllEnvConfig() || {};
   // When globalConfigManager is undefined, configMap = {}
   ```

4. With `configMap = {}`, no model is configured, triggering the error.

## Suggested Fix

**Option 1**: The condition should not trigger new `ModelConfigManager` for `createOpenAIClient` alone:

```javascript
const hasCustomConfig = opts?.modelConfig;  // Remove || opts?.createOpenAIClient
this.modelConfigManager = hasCustomConfig ? new ModelConfigManager(opts?.modelConfig, opts?.createOpenAIClient) : globalModelConfigManager;
```

**Option 2**: If creating a new instance, automatically register `globalConfigManager`:

```javascript
const hasCustomConfig = opts?.modelConfig || opts?.createOpenAIClient;
this.modelConfigManager = hasCustomConfig ? new ModelConfigManager(opts?.modelConfig, opts?.createOpenAIClient) : globalModelConfigManager;
if (!this.modelConfigManager.globalConfigManager) {
  this.modelConfigManager.registerGlobalConfigManager(globalConfigManager);
}
```

## Impact

This prevents users from using `createOpenAIClient` to integrate Langfuse, LangSmith, or other OpenAI wrapper libraries, which is a common observability requirement.

## Workaround

None available without modifying Midscene source code or forking the package.

### Reproduce link

/

### Reproduce Steps

```typescript
import { PlaywrightAgent } from '@midscene/web/playwright';

const agent = new PlaywrightAgent(page, {
  groupName: 'test',
  reportFileName: 'test-report',
  cache: { id: 'test-cache' },
  // Even an empty/no-op hook breaks the configuration
  createOpenAIClient: async (baseOpenAI) => {
    return baseOpenAI;
  },
});
```

linhongyu510:
## Summary

- preserve global model configuration when `createOpenAIClient` is provided without an explicit `modelConfig`
- expose the agent-scoped client factory through `agent.modelConfigManager`, including Playground describe and locate runtimes
- keep custom client factories isolated between agents that share global model values

## Root cause

`Agent` treated `createOpenAIClient` alone as a custom model configuration and created an isolated `ModelConfigManager`. Because that manager had neither an explicit `modelConfig` nor a registered `GlobalConfigManager`, it resolved an empty configuration and reported that `MIDSCENE_MODEL_NAME` was missing.

Simply reusing the global manager would fix model resolution but would drop the agent-scoped factory from public `modelConfigManager` consumers. Playground builds describe and locate runtimes through that public manager, so observability wrappers such as LangSmith or Langfuse would be silently skipped on those paths.

## Solution

Factory-only agents now use an `AgentScopedModelConfigManager` that delegates model values to `globalModelConfigManager` and returns a copied configuration with the current agent's `createOpenAIClient`.

This keeps global overrides and environment updates dynamic, avoids mutating shared cached configurations, and preserves independent factories for multiple agents.

Tests use stubbed environment variables with explicit cache clearing so global configuration state is restored after every case.

## Validation

- `pnpm --filter @midscene/core test`
  - 150 test files passed
  - 1,564 tests passed, 8 skipped
- `pnpm build`
  - 28 projects built successfully
- `pnpm type-check:tests`
- `pnpm run lint`

Closes #3079.

### DIFF
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -328,9 +328,11 @@ export class Agent<InterfaceType extends AbstractInterface = AbstractInterface>
   }
 
   private resolveModelRuntime(intent: TIntent): ModelRuntime {
-    const runtime = getModelRuntime(
-      this.modelConfigManager.getModelConfig(intent),
-    );
+    const modelConfig: IModelConfig = {
+      ...this.modelConfigManager.getModelConfig(intent),
+      createOpenAIClient: this.opts.createOpenAIClient,
+    };
+    const runtime = getModelRuntime(modelConfig);
     return {
       ...runtime,
       onUsage: (usage) => {
@@ -422,11 +424,10 @@ export class Agent<InterfaceType extends AbstractInterface = AbstractInterface>
         `opts.modelConfig must be a plain object map of env keys to values, but got ${typeof opts?.modelConfig}`,
       );
     }
-    // Create ModelConfigManager if modelConfig or createOpenAIClient is provided
-    // Otherwise, use the global config manager
-    const hasCustomConfig = opts?.modelConfig || opts?.createOpenAIClient;
-    this.modelConfigManager = hasCustomConfig
-      ? new ModelConfigManager(opts?.modelConfig, opts?.createOpenAIClient)
+    // Explicit modelConfig is isolated from global configuration.
+    // A custom client factory alone still uses the global model values.
+    this.modelConfigManager = opts?.modelConfig
+      ? new ModelConfigManager(opts.modelConfig, opts.createOpenAIClient)
       : globalModelConfigManager;
 
     this.onTaskStartTip = this.opts.onTaskStartTip;
--- a/packages/core/src/agent/agent.ts
+++ b/packages/core/src/agent/agent.ts
@@ -69,6 +69,7 @@ import {
   observationArtifactAdapterSymbol,
 } from '@midscene/shared/agent-tools/observation-artifact';
 import {
+  type CreateOpenAIClientFn,
   type IModelConfig,
   MIDSCENE_REPLANNING_CYCLE_LIMIT,
   ModelConfigManager,
@@ -119,6 +120,35 @@ import {
 
 const debug = getDebug('agent');
 const warn = getDebug('agent', { console: true });
+
+class AgentScopedModelConfigManager extends ModelConfigManager {
+  constructor(
+    private readonly baseManager: ModelConfigManager,
+    private readonly createOpenAIClient?: CreateOpenAIClientFn,
+  ) {
+    super();
+  }
+
+  override getModelConfig(intent: TIntent): IModelConfig {
+    return {
+      ...this.baseManager.getModelConfig(intent),
+      createOpenAIClient: this.createOpenAIClient,
+    };
+  }
+
+  override getUploadTestServerUrl(): string | undefined {
+    return this.baseManager.getUploadTestServerUrl();
+  }
+
+  override throwErrorIfNonVLModel() {
+    const modelConfig = this.getModelConfig('default');
+    if (!modelConfig.modelFamily) {
+      throw new Error(
+        'MIDSCENE_MODEL_FAMILY is not set to a multimodal model with UI localization, so element localization cannot be achieved. Check your model configuration. See https://midscenejs.com/model-strategy.html',
+      );
+    }
+  }
+}
 
 export type AiActOptions = {
   cacheable?: boolean;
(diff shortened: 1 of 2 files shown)


EAGzzyCSL:
Thanks for the fix! I think wrapping `globalModelConfigManager` in the Agent constructor is the right approach here. It preserves the global model configuration while making the agent-specific factory available to all consumers of `agent.modelConfigManager`, including Playground.

Could we simplify the implementation slightly?

- Keep the config copy and factory injection in `AgentScopedModelConfigManager.getModelConfig()`.
- Remove the additional shallow copy in `resolveModelRuntime()` and pass `this.modelConfigManager.getModelConfig(intent)` directly to `getModelRuntime()`.
- Remove the `getUploadTestServerUrl()` and `throwErrorIfNonVLModel()` overrides. Their inherited implementations already call `this.getModelConfig()`, so they should work with the wrapper.

This keeps the factory injection in one place without changing the behavior of the fix.

linhongyu510:
Addressed the simplification in `283c3fd0`: `AgentScopedModelConfigManager` now only overrides `getModelConfig()`, while inherited upload-server and non-VL validation methods reuse that injected config; `resolveModelRuntime()` also passes the manager result directly without another shallow copy. Verification: focused custom-model suite 21/21 passed, full Biome lint checked 1,627 files, `nx build @midscene/core` passed, and `git diff --check` passed.

EAGzzyCSL (approved):

PR #3156, opened 2026-09-17
fix(core): add quick mappings for scroll and swipe directions

EAGzzyCSL:
## Summary

Models can identify that content below the viewport needs to be revealed but still emit `Scroll.direction: up`. Add explicit intent-to-direction quick mappings to the Scroll and Swipe parameter descriptions, including finger/pointer and content movement for Swipe. Update the corresponding assertion and inline snapshot.

## Validation

- `pnpm run lint`
- `pnpm --filter @midscene/core test:u tests/unit-test/prompt/prompt.test.ts tests/unit-test/prompt/planning/action-description.test.ts -u` — 60 tests passed
- Ran `npx --no-install tsx my-android.ts` serially 10 times before and 10 times after rebuilding core, using `qwen3.7-plus` with the same script and identical screenshots. First-plan direction errors were 2/10 before and 1/10 after; errors across all planning rounds were 8/40 and 4/40. This small sample does not establish a reliable reduction in error rate. All runs reached the replanning limit with unchanged screenshots, so this measures direction selection rather than task completion.


EAGzzyCSL:
beta：https://github.com/web-infra-dev/midscene/actions/runs/35194577597

